PR #16, opened 2015-03-17
Add menu bar with catalog entries

wandernauta:
I added a quick menu bar to the GUI to show how the catalog can be used.
Clicking the elements doesn't actually *do* anything yet, but it does show up.

@TheMSB @Kristel-92 @i3anaan Would this be useful to put in master? Please
advise. Thanks.


TheMSB:
It might be useful in some later debug context, I don't think it's wise to merge this into master though, especially considering we will not be using menu bars of any kind for the final product.

wandernauta:
Understood, closing.

PR #20, opened 2015-03-23
Refactoring of FunctionBlock

TheMSB:
Refactored FunctionBlock in order to make feature expansion easier.
FunctionBlock now extends Block and certain methods have subsequently been moved from FunctionBlock to Block.

newInstance of FunctionBlock has been slightly rewritten and a second version of newInstance has been introduced to enable the instantiation of FunctionBlocks with initialised names.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
@@ -15,8 +15,8 @@ public class Block extends StackPane {
 	 * Set the selected boolean state of this Block
 	 * @param selectedState
 	 */
-		public void setSelected(boolean selectedState){
-			//TODO If another object is selected then deselect it first!!
-			isSelected = selectedState;
-		}
+	public void setSelected(boolean selectedState) {
+		//TODO If another object is selected then deselect it first!!
+		isSelected = selectedState;
+	}
 }
--- a/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
@@ -12,11 +12,11 @@ public class Block extends StackPane {
 	private boolean isSelected = false;
 	
 	/**
-	 * Set the selected state of this Block
-	 * @param Selected False/True
+	 * Set the selected boolean state of this Block
+	 * @param selectedState
 	 */
-		public void setSelected(boolean bool){
+		public void setSelected(boolean selectedState){
 			//TODO If another object is selected then deselect it first!!
-			isSelected = bool;
+			isSelected = selectedState;
 		}
 }
--- a/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
@@ -1,0 +1,22 @@
+package nl.utwente.group10.ui.components;
+
+import javafx.scene.layout.StackPane;
+
+/**
+ * Base UI Component that other visual elements will extend from.
+ * If common functionality is found it should be refactored to here.
+ */
+public class Block extends StackPane {
+
+	/** Selected state of this Block*/
+	private boolean isSelected = false;
+	
+	/**
+	 * Set the selected state of this Block
+	 * @param Selected False/True
+	 */
+		public void setSelected(boolean bool){
+			//TODO If another object is selected then deselect it first!!
+			isSelected = bool;
+		}
+}
--- a/Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
@@ -43,6 +43,20 @@ public class FunctionBlock extends Block{
 	}
 	
 	/**
+	 * Method that creates a newInstance of this class along with it's visual representation
+	 * @param the number of arguments this FunctionBlock can hold
+	 * @param the name of this FunctionBlock
+	 * @return a new instance of this class
+	 * @throws IOException
+	 */
+	public static FunctionBlock newInstance(int numberOfArguments, String name) throws IOException{
+		FunctionBlock functionBlock = newInstance(numberOfArguments);
+		functionBlock.setName(name);
+		
+		return functionBlock;
+	}
+	
+	/**
 	 * Executes this FunctionBlock and returns the output as a String
 	 * @return Output of the Function
 	 */
--- a/Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
@@ -18,11 +18,11 @@ import javafx.scene.shape.Rectangle;
  * represents a Haskell function together with it's arguments and
  * visual representation.
  */
-public class FunctionBlock extends StackPane{
-	//The arguments this FunctionBlock holds
+public class FunctionBlock extends Block{
+	/**The arguments this FunctionBlock holds.**/
 	private String[] arguments;
+	/** The name of this Function.**/
 	private String functionName;
-	private boolean isSelected = false;
 	
 	/**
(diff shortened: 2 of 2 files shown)


Code/src/main/java/nl/utwente/group10/ui/components/Block.java
```diff
@@ -0,0 +1,22 @@
...
+public class Block extends StackPane {
+
+	/** Selected state of this Block*/
+	private boolean isSelected = false;
+	
+	/**
+	 * Set the selected boolean state of this Block
+	 * @param selectedState
+	 */
+		public void setSelected(boolean selectedState){
```
wandernauta: Nitpick: this seems to be indented just one level further than necessary.

Code/src/main/java/nl/utwente/group10/ui/components/Block.java
```diff
@@ -0,0 +1,22 @@
...
+			//TODO If another object is selected then deselect it first!!
```
wandernauta: Is this the right place to do this? Would you expect foo.setSelected(true) to affect bar.isSelected() even if foo and bar appear to be unrelated?

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -18,11 +18,11 @@
  * represents a Haskell function together with it's arguments and
  * visual representation.
  */
-public class FunctionBlock extends StackPane{
-	//The arguments this FunctionBlock holds
+public class FunctionBlock extends Block{
```
wandernauta: Missing space before {.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -31,32 +31,54 @@
...
-		//TODO Prettier resource loading
-		//TODO better factory
-		FunctionBlock functionBlock = null;
+		FunctionBlock functionBlock;
 		try{
 			functionBlock = (FunctionBlock) FXMLLoader.load(Main.class.getResource("/ui/FunctionBlock.fxml"), null, new TactileBuilderFactory());
 			functionBlock.initializeArguments(numberOfArguments);
 		}catch(IOException e){
 			e.printStackTrace();
+			functionBlock = null;
```
wandernauta: I don't like this catch. All this will do is cause NullPointerExceptions down the line, so why catch it at all?

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -31,32 +31,54 @@
...
-	//TODO To be implemented method that will parse code into Haskell interface
-	// and returns the result
+	/**
+	 * Method that creates a newInstance of this class along with it's visual representation
+	 * @param the number of arguments this FunctionBlock can hold
+	 * @param the name of this FunctionBlock
+	 * @return a new instance of this class
+	 * @throws IOException
+	 */
+	public static FunctionBlock newInstance(int numberOfArguments, String name) throws IOException{
```
wandernauta: Missing space.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -31,32 +31,54 @@
...
 	public String executeMethod(){		
 		return new String("DEBUG-OUTPUT");
 	}
 	
+	/**
+	 * Nest another Node object within this FunctionBlock
+	 * @param node to nest
+	 */
 	public void nest(Node node){
 		Pane nestSPace = (Pane) this.lookup("#nest_space");
```
wandernauta: Is it a `nestSPace`, `nestSpace`, or `nestPane`?

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -31,32 +31,54 @@
...
 		((Label) this.lookup("#label_function_name")).setText("Higher order function");
 		nestSPace.getChildren().add(node);
 	}
 	
-	//Private method to initialize the argument fields for this function block.
-	//All arguments will be stored as Strings.
+	/**
+	 * Private method to initialize the argument fields for this function block.
```
wandernauta: Nitpick: Documentation is not very useful.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -31,32 +31,54 @@
...
+	 * All arguments will be stored as Strings.
```
wandernauta: What would those Strings look like, exactly?

wandernauta:
Big up for the refactoring, great idea. Some minor nitpicks and questions above, other than that :+1:

Code/src/main/java/nl/utwente/group10/ui/components/Block.java
TheMSB: Any Block within the UI can be selected, that may be extended beyond just FunctionBlocks. (think TypeBlock or OutputBlock), since they would all share this function I see no need to implement it multiple times in each respective extended class.

wandernauta: Very true, but I was specifically referring to the TODO: would the setSelected method deselect another object, and how would it find that other object? Or should this be in the documentation, i.e. should someone else who calls this method call setSelected on the other object first?

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
TheMSB: Catch was already in place, I just placed the null initialisation in a more appropriate place.
If so desired the try catch can be removed.

wandernauta: Yes please :relaxed:

TheMSB:
Committed code styling fixes, will rework other code after discussion.

wandernauta:
Sounds good!

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
TheMSB: Fixed in next commit.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
TheMSB: Good question, as I did not develop this piece of code but just provided documentation from what I understood of it. I believe the name should be nestSpace and as such will be fixed in next commit.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
TheMSB: Strings will contain exactly what is entered though the user interface into the argument field, Since there is no way to actually set them as of yet but the need to store them is clear the initialisation has already been done. Next commit will contain a less obscure javaDoc.

Code/src/main/java/nl/utwente/group10/ui/components/Block.java
TheMSB: If an object is selected through the 'tap' gesture, other objects would have to be un-selected since the 'tap' gesture means: select THIS specific object. Hence the TODO. It is however still undetermined how group selections will be implemented exactly.

TheMSB:
Commit second batch of changes, all issues *should* be resolved.

wandernauta:
Looks good to me, will merge when Travis is green.

Code/src/main/java/nl/utwente/group10/ui/components/Block.java
wandernauta: Okay, sounds reasonable.

wandernauta:
Merged, pleasure doing business! :shipit:

PR #22, opened 2015-03-26
UI Extension of Block features and Classes

TheMSB:
Created ValueBlocks to have base values to work off of with FunctionBlocks.
Created OutputBlocks to display output values coming from Value and FunctionBlocks(technically any Block that serves output). OutputBlocks can also relay the output after displaying to another Block.

fxml and style files have been included to reflect changes.


Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.IOException;
+
+/**
+ * OutputBlock is an extension of Block that only provides a display of the input
+ * it receives through it's inputAnchor.
+ * The input will be rendered visually on the Block.
+ * OutputBlock can be empty and contain no value at all, the value can be altered at any time
+ * by providing a different input source using a Connection.
+ */
+public class OutputBlock extends Block {
```
wandernauta: Name might be slightly confusing: this is the only subclass of Block the does *not* have any output itself.

TheMSB: But it can, OutputBlock is capable of rerouting the input it receives and use it as output.

TheMSB: While the actual method in charge if returning output is non-existent in the commits as of yet I intend to add it once ui-connections has finished. If so desired the method can be included in an additional commit now but it is likely to be revised later.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
```diff
@@ -0,0 +1,67 @@
...
+	
+	/** The input this Block is receiving.**/
+	private String input;
```
wandernauta: This seems to be a push-based model: you apparently expect functions to *push* their own results into all connected output blocks. This is reasonable, but it might be easier to have a pull-based model, where blocks like these *pull* their values from their 'parents' when they need them. This makes invalidation slightly more straightforward, and means that FunctionBlock does not need to know about OutputBlock.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
```diff
@@ -0,0 +1,67 @@
...
+	/** Determines if this Block serves output.**/
+	private boolean output;
+	
+	/**
+	 * Creates a new instance of OutputBlock should only ever be called
+	 * internally by the class, for instantiation see other newInstance method.
```
wandernauta: This contract has already been enforced by the method being private.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
```diff
@@ -0,0 +1,67 @@
...
+    private static OutputBlock newInstance() throws IOException {
+        OutputBlock block = FXMLLoader.load(OutputBlock.class.getResource("/ui/OutputBlock.fxml"), null, new TactileBuilderFactory());
+        ((Label) block.lookup("#label_value")).setText("Connect Input");
+        return block;
+    }
+    
+    /**
+     * Creates a new OutputBlock instance, servesThroughput is a boolean value to determine
+     * if the newly created instance should initially put input through as output in addition to
+     * displaying the value.
```
wandernauta: Why not always enable throughput?

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Method to enable/disable if this OutputBlock serves output
+     * in addition to displaying it.
+     * @param enabledState
+     */
+    public void enableOutput(boolean enabledState) {
+    	output = enabledState;
+    }
+    
+}
```
wandernauta: Newline end of file, also weird mix of tabs and spaces in this file

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
wandernauta: Why not call it a `DisplayBlock` or something like that, if that's its primary use?

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
TheMSB: Indeed an oversight, noted and fixed.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
TheMSB: Design decision, naturally an OutputBlock with throughput enabled does indeed have the same functionality. Disabling it reduces visual children. In hindsight I have no other justification for this, class can indeed be simplified.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
TheMSB: Fixed.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
```diff
@@ -0,0 +1,67 @@
...
+	
+	/**
+	 * Creates a new instance of OutputBlock should only ever be called
+	 * internally by the class, for instantiation see other newInstance method.
+	 * @return new OutputBlock instance
+	 * @throws IOException
+	 */
+    private static OutputBlock newInstance() throws IOException {
+        OutputBlock block = FXMLLoader.load(OutputBlock.class.getResource("/ui/OutputBlock.fxml"), null, new TactileBuilderFactory());
+        ((Label) block.lookup("#label_value")).setText("Connect Input");
```
wandernauta: Default text could probably be put in the FXML.

Code/src/main/java/nl/utwente/group10/ui/components/ValueBlock.java
```diff
@@ -0,0 +1,32 @@
...
+	 * @param value of this ValueBlock
+	 * @return new ValueBlock instance
+	 * @throws IOException
+	 */
+    public static ValueBlock newInstance(String value) throws IOException {
+        ValueBlock block = FXMLLoader.load(ValueBlock.class.getResource("/ui/ValueBlock.fxml"), null, new TactileBuilderFactory());
+        ((Label) block.lookup("#label_value")).setText(value);
+        return block;
+    }
+}
```
wandernauta: Newline end of file

Code/src/main/java/nl/utwente/group10/ui/components/ValueBlock.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * ValueBlock is an extension of Block that contains only a value
+ * and does not accept input of any kind.
+ * A single output source will be generated in order to connect
+ * a ValueBlock to another Block.
+ * 
+ * Extensions of ValueBlock should never accept inputs, if so desired
+ * Block should be extended instead.
+ */
+public class ValueBlock extends Block {
```
wandernauta: OutputBlock expects results to be pushed in, yet ValueBlock never pushes.

Code/src/main/resources/ui/FunctionBlock.fxml
```diff
@@ -4,7 +4,7 @@
 <?import java.net.*?>
 
 
-<FunctionBlock xmlns:fx="http://javafx.com/fxml/1" styleClass="functionBlock">
+<FunctionBlock xmlns:fx="http://javafx.com/fxml/1" styleClass="block">
 	<BorderPane id="foreground" minHeight="90" minWidth = "180">
```
wandernauta: Whitespace around =.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
TheMSB: Method has been added with new commit.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
TheMSB: Can be implemented when ui-connections has been completed. Blocks will then be able to pull values from Anchor objects.

Code/src/main/java/nl/utwente/group10/ui/components/ValueBlock.java
TheMSB: You want a trailing newline?

Code/src/main/java/nl/utwente/group10/ui/components/ValueBlock.java
TheMSB: Push/Pull behaviour will be modified once ui-connections is finished, suggestion of push behaviour has been removed from OutputBlock.

Code/src/main/resources/ui/FunctionBlock.fxml
TheMSB: fixed.

Code/src/main/java/nl/utwente/group10/ui/components/OutputBlock.java
TheMSB: Fixed.

Code/src/main/java/nl/utwente/group10/ui/components/ValueBlock.java
wandernauta: Yes, for every file. Eclipse should have a setting to enforce this. ([Here's why if you're curious.](http://stackoverflow.com/a/14637072))

wandernauta:
General notes:

- The classes you introduce do not seem to be used anywhere. This makes it a bit hard to see if they're actually working. 
- Deciding on whether to push values from parents or pull them from children is actually quite a big difference on a structural level, you might want to discuss this with others. From what I can see it looks a bit like the PR in its current state does neither.

wandernauta:
There's some duplicated code in the JavaFX-facing tests, if at all possible you could refactor that (and add something that closes the JavaFX threads when you're done).

wandernauta:
Merged, thanks!

PR #23, opened 2015-03-26
function block select & drag gesture

Kristel-92:
CustomGestureEvent wordt niet meer gebruikt. In plaats hiervan is dit nu UIEvent. De code is echter hetzelfde gebleven. Verder vangt (Function)Block nu ook events af en houdt een (Function)Block bij in welke CustomUIPane deze zit i.v.m. het selecteren van ander (Function)Blocks.
Een (Function)Block kan nu geselecteerd worden door er op te klikken. Bij het klikken op een ander (Function)Block vervalt een eventuele andere selectie

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
@@ -9,13 +9,9 @@ import nl.utwente.group10.ui.gestures.GestureCallBack;
 public class CustomUIPane extends TactilePane implements GestureCallBack {
 
 	public CustomUIPane() {
-		CustomGesture cg = new CustomGesture(this, this);
 	}
 
 	@Override
 	public void handleCustomEvent(UIEvent event) {
-		EventType<UIEvent> eventType = (EventType<UIEvent>) event
-				.getEventType();
-	
 	}
 }
--- a/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
@@ -3,7 +3,7 @@ package nl.utwente.group10.ui;
 import javafx.event.EventType;
 import nl.utwente.ewi.caes.tactilefx.control.TactilePane;
 import nl.utwente.group10.ui.gestures.CustomGesture;
-import nl.utwente.group10.ui.gestures.CustomGestureEvent;
+import nl.utwente.group10.ui.gestures.UIEvent;
 import nl.utwente.group10.ui.gestures.GestureCallBack;
 
 public class CustomUIPane extends TactilePane implements GestureCallBack {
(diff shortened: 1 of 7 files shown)


Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
```diff
@@ -13,18 +13,9 @@ public CustomUIPane() {
 	}
 
 	@Override
-	public void handleCustomEvent(CustomGestureEvent event) {
-		EventType<CustomGestureEvent> eventType = (EventType<CustomGestureEvent>) event
+	public void handleCustomEvent(UIEvent event) {
+		EventType<UIEvent> eventType = (EventType<UIEvent>) event
```
wandernauta: It seems this variable is not used, can it be deleted?

Code/src/main/java/nl/utwente/group10/ui/components/Block.java
```diff
@@ -19,4 +26,23 @@ public void setSelected(boolean selectedState) {
 		//TODO If another object is selected then deselect it first!!
 		isSelected = selectedState;
 	}
+	
+	@Override
+	public void handleCustomEvent(UIEvent event) {
+		EventType<UIEvent> eventType = (EventType<UIEvent>) event
```
wandernauta: Is this cast necessary? I'm not sure, but you might get away with just calling this an `EventType` instead of an `EventType<UIEvent>`.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -23,17 +28,20 @@
...
+	private static CustomGesture cg;
+		
 	/**
 	 * Method that creates a newInstance of this class along with it's visual representation
 	 * @param the number of arguments this FunctionBlock can hold
 	 * @return a new instance of this class
 	 * @throws IOException
 	 */
-	public static FunctionBlock newInstance(int numberOfArguments) throws IOException {
+	public static FunctionBlock newInstance(int numberOfArguments, CustomUIPane pane) throws IOException {
```
wandernauta: If possible, update the JavaDoc to mention what kind of pane you expect and why it needs to be passed.

wandernauta:
Looks very reasonable, other than a few minor nitpicks. :+1: if those can be fixed.

Code/src/main/java/nl/utwente/group10/ui/components/Block.java
```diff
@@ -1,16 +1,23 @@
...
  * If common functionality is found it should be refactored to here.
  */
-public class Block extends StackPane {
+public class Block extends StackPane implements GestureCallBack {
 
 	/** Selected state of this Block*/
 	private boolean isSelected = false;
 	
+	protected static CustomUIPane cup;
+
```
TheMSB: Question:
What is this for? I cannot tell why the Block needs to store the CustomUIPane.
Request:
Add javaDoc, it could clarify the usage of some code.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
TheMSB: Suggestion:
Isn't it possible to get the CustomUIPane by looking up the FunctionBlock's parent?
It seems like there is no need to pass along and store it in a custom variable.

wandernauta: Nicely spotted!

I don't think just taking the parent would be enough, though. At the very least, and I might be wrong in this, I think you'd have to walk the tree upwards, as I believe nodes (such as function blocks) can theoretically be nested. On the other hand, that case is not supported with the current approach either, which requires all blocks to be direct descendants of the pane.

To make a long story short: if all blocks are and will always remain direct descendants of Pane, go with @TheMSB's suggestion; if not, have Blocks (or FunctionBlocks) keep a reference to some global state like you do here, and perhaps add a method to find everything that looks like a block.

Alternatively, and this is the solution I like most even though it is probably a bit more work, have some observable global state object (perhaps Main?) [reference](https://docs.oracle.com/javafx/2/api/javafx/beans/value/ObservableObjectValue.html) a single Block that is *the* selected block, then push out events to every Block when that reference changes. Blocks would then each check if they're the selected block, and update their own state accordingly. This ensures that there is always just one (or always 0 or 1 if you wrap it in an Optional) object selected and keeps mutable state as simple as possible.

TheMSB: @wandernauta , I am not entirely sure on the specifics of javaFX but wouldn't it be possible to "look, up" in the tree until it encounters a parent with specific name? This could greatly simplify the structure.

wandernauta: Yes, that would be the 'walking the tree upwards' approach I think. All of the four approaches are reasonable, really, it's just a matter of preference.

Kristel-92:
requests done

wandernauta:
Merged, thanks!

PR #25, opened 2015-03-26
Refactor type system to use Hindley-Milner type inference

jjkester:
Introduces very large changes under the hood, complete with API breaking changes. All code that was present at the creation of this branch has been ported to the new API, including the type parser.

Major changes:
- Experimental support for type classes dropped until further notice.a
- API changes - code using classes from the ```haskell``` package should be checked.
- Using a well-known, mathematically proven type inference (although the implementation is not necessarily mathematically correct).
- Better environment management
- XML-based catalog for supported Haskell functions


Code/src/main/java/nl/utwente/group10/haskell/catalog/Entry.java
```diff
@@ -1,40 +1,68 @@
...
 
-    public EnvFunc asEnvFunc() {
-        TypeBuilder tb = new TypeBuilder();
-        return new EnvFunc(this.name, tb.buildFuncT(this.signature));
+    /**
+     * Parses and returns the Type of the function in this Entry.
+     * @return The Type of this Entry.
+     */
+    public final Type getType() {
+        return null; // TODO Let this return the type of this Entry when the type parser is fixed.
```
wandernauta: Might give this a shot, if the type parser is indeed fixed?

Code/src/main/java/nl/utwente/group10/haskell/env/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Provides an environment for Haskell.
+ */
+package nl.utwente.group10.haskell.env;
```
wandernauta: Newline end-of-file missing here and on some other new files. 

(Hint: File > Settings > Editor > General > Other > Ensure line feed at end of file on Save)

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
```diff
@@ -1,66 +1,60 @@
...
+import nl.utwente.group10.haskell.hindley.GenSet;
+import nl.utwente.group10.haskell.hindley.HindleyMilner;
 import nl.utwente.group10.haskell.type.FuncT;
 import nl.utwente.group10.haskell.type.Type;
 
-import java.util.Arrays;
-
 /**
- * A type of expression that is a lazy application of a function with certain arguments.
+ * Lazy application of an argument to a function.
```
wandernauta: What is lazy about an Apply, specifically?

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
```diff
@@ -1,66 +1,60 @@
...
+     * The argument to apply.
      */
-    private final Expr[] args;
+    private final Expr arg;
 
     /**
-     * Constructs a new function application object. The resulting type will be calculated based on the given arguments.
-     * The type of the arguments is validated, and an exception is thrown when one of the arguments is incorrect.
+     * Applies an argument to a function to produce a new expression. The application is lazy, the type needs to be
+     * analyzed before it can be resolved.
```
wandernauta: Is whoever calls this constructor responsible for doing the analyzing and resolving?

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
```diff
@@ -1,66 +1,60 @@
...
+        return String.format("(%s %s)", this.func, this.arg);
     }
 
     @Override
     public final String toString() {
-        return "Apply{" +
-                "func=" + this.func +
-                ", args=" + Arrays.toString(this.args) +
-                '}';
+        return String.format("(%s %s)", this.func, this.arg);
```
wandernauta: Seeing as this is now identical to `toHaskell`, I'd just call that.

Code/src/main/java/nl/utwente/group10/haskell/expr/Expr.java
```diff
@@ -1,30 +1,26 @@
...
-    protected Expr(final Type type) {
-        this.type = type;
-    }
-
-    /**
-     * @return The type of this Haskell expression.
+     * Analyzes the type tree and resolves the type for this usage of this expression.
+     *
+     * @param env The current Haskell environment.
+     * @param genSet TODO Find out what this does...
```
wandernauta: TODO noted. Still haven't got the foggiest idea myself :grimacing:

Code/src/main/java/nl/utwente/group10/haskell/expr/Expr.java
```diff
@@ -1,30 +1,26 @@
...
+     * @return The type for this usage of this expression.
+     * @throws HaskellTypeError The type tree contains an application of an incompatible type.
      */
-    public final Type getType() {
-        return this.type;
-    }
+    public abstract Type analyze(final Env env, final GenSet genSet) throws HaskellTypeError;
```
wandernauta: Does `final genSet` mean we can't modify the `genSet` or does it mean we can't replace the `genSet` with a different one? We might need to do the first.

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
```diff
@@ -0,0 +1,42 @@
...
+import nl.utwente.group10.haskell.exceptions.HaskellTypeError;
+import nl.utwente.group10.haskell.hindley.GenSet;
+import nl.utwente.group10.haskell.type.Type;
+
+/**
+ * Expression contained in the environment.
+ */
+public class Ident extends Expr {
+    /**
+     * Name of this identity.
```
wandernauta: Not really, `Ident` is short for *identifier* not *identity*, and `name` is a bit of a misnomer: it's really the identifier itself.

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
```diff
@@ -0,0 +1,42 @@
+package nl.utwente.group10.haskell.expr;
+
+import nl.utwente.group10.haskell.env.Env;
+import nl.utwente.group10.haskell.exceptions.HaskellTypeError;
+import nl.utwente.group10.haskell.hindley.GenSet;
+import nl.utwente.group10.haskell.type.Type;
+
+/**
+ * Expression contained in the environment.
```
wandernauta: Isn't it more of a *reference* to an expression (function) contained in the environment?

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public final String toHaskell() {
+        return this.name;
+    }
+
+    @Override
+    public final String toString() {
+        return this.name;
```
wandernauta: Might want to consider returning `toHaskell` here. I'd be perfectly okay with not bothering.

Code/src/main/java/nl/utwente/group10/haskell/hindley/HindleyMilner.java
```diff
@@ -0,0 +1,86 @@
...
+        } else if (a instanceof ConstT && b instanceof ConstT) {
+            // Example: we have to unify Int and Int.
+
+            final ConstT ao = (ConstT) a;
+            final ConstT bo = (ConstT) b;
+
+            // If the constructor doesn't match, give up right away.
+            // Example: trying to unify String and Int.
+            if (!ao.getConstructor().equals(bo.getConstructor())) {
+                throw new HaskellTypeError(String.format("Sadly, these types are not compatible: %s ≠ %s", a, b), context);
```
wandernauta: For a next PR, it might be worthwhile looking into HaskellTypeError and make it a bit more informative for frontend: for example, if we can somehow deduce that it's the seventh argument of such-and-such apply, frontend would probably love to know that.

Code/src/main/java/nl/utwente/group10/haskell/hindley/HindleyMilner.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Creates and returns a new {@code VarT} instance with a unique identifier.
+     * @return A new variable type.
+     */
+    public static VarT makeVariable() {
+        final String name;
+
+        name = HindleyMilner.tvOffset <= 25
+                ? String.valueOf((char) ('α' + HindleyMilner.tvOffset))
+                : Integer.toString(HindleyMilner.tvOffset);
```
wandernauta: Normally not a big fan of ternary operators, but this is pretty neat.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -1,35 +1,73 @@
...
+    private final String constructor;
+
+    /**
+     * The types of the arguments for this type.
+     */
+    private final Type[] args;
+
+    /**
+     * @param constructor The constructor for this constant type.
+     * @param args The types of the arguments that this type accepts.
```
wandernauta: Might want to add '...if any' - quite a lot of types have nullary type constructors.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -1,35 +1,73 @@
...
-        this.name = name;
+    public final Type[] getArgs() {
+        return this.args;
     }
 
     @Override
-    public final boolean compatibleWith(final Type other) {
-        return this.equals(other);
+    public final Type prune() {
+
```
wandernauta: Superfluous newline.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -1,35 +1,73 @@
...
-    public final String toString() {
-        return "ConstT{" +
-                "name='" + this.name + "'" +
-                "}";
+    public String toString() {
+        if (this.args.length > 0) {
+            return String.format("(%s %s)", this.constructor, Joiner.on(' ').join(this.args));
+        } else {
+            return this.constructor;
+        }
```
wandernauta: What's the difference between this `toString` and the `toHaskellType` above, apart from the parentheses?

Code/src/main/java/nl/utwente/group10/haskell/type/TupleT.java
```diff
@@ -1,68 +1,32 @@
 package nl.utwente.group10.haskell.type;
 
+import com.google.common.base.Joiner;
```
wandernauta: Is this import still used?

Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
```diff
@@ -2,45 +2,29 @@
...
+import java.util.logging.Logger;
+
 /**
  * Abstract class for Haskell types. Provides an interface for common methods.
  *
  * Types should be immutable. Also, a Type subclass instance that represents the same in Haskell as another instance of
  * the same class should be equal, that is {@code this.equals(that)} should return {@code true}.
  */
-public abstract class Type extends HaskellObject implements Comparable<Type> {
+public abstract class Type extends HaskellObject {
```
wandernauta: The JavaDoc above might no longer be true: unify does some horrible, horrible things to types, which probably makes them no longer immutable, and the instance/equals stuff I'm not sure about either.

Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
```diff
@@ -2,45 +2,29 @@
...
     /**
-     * Tests whether the given type is usable within the context of this type.
-     *
-     * Examples:
-     * - The `Integer` type is valid within the type class `Num`
-     * - Every type is valid within a {@code VarT} type.
-     *
-     * @param other The type to test.
-     * @return Whether the given type is usable in place of this type.
+     * Prunes the type subtree and returns the resolved type for this type. The resolved type can be this type itself.
```
wandernauta: This JavaDoc could use a nice example of what prune does, exactly. I'm half-sure I forgot it already. (Perhaps also call it instanced or maybe even inferred, not resolved.)

Code/src/test/java/nl/utwente/group10/haskell/expr/ApplyTest.java
```diff
@@ -1,64 +1,95 @@
...
-import nl.utwente.group10.haskell.exceptions.HaskellException;
+import nl.utwente.group10.haskell.env.Env;
+import nl.utwente.group10.haskell.exceptions.HaskellTypeError;
+import nl.utwente.group10.haskell.hindley.GenSet;
 import nl.utwente.group10.haskell.type.*;
 import org.junit.Before;
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
wandernauta: I'll allow it.

Code/src/test/java/nl/utwente/group10/haskell/expr/ApplyTest.java
```diff
@@ -1,64 +1,95 @@
...
-    private Type T_b;
-    private Type T_c;
+    private final Type alpha = new VarT("a");
+    private final Type beta = new VarT("b");
+    private final Type gamma = new VarT("c");
+    private final Type alphaList = new ListT(this.alpha);
+    private final Type betaList = new ListT(this.beta);
+    private final Type integer = new ConstT("Int");
+    private final Type integerList = new ListT(this.integer);
+    private final Type string = new ConstT("Str");
```
wandernauta: The Haskell type is really `String`, which is an alias for `[Char]` - a different example might be clearer.

Code/src/test/java/nl/utwente/group10/haskell/expr/ApplyTest.java
```diff
@@ -1,64 +1,95 @@
...
+        this.env.put("zip", new FuncT(this.alphaList, new FuncT(this.betaList, new ListT(new TupleT(this.alpha, this.beta)))));
+        this.env.put("lcm", new FuncT(this.alpha, new FuncT(this.alpha, this.alpha)));
+    }
+
+    @Test
+    public final void testId() throws HaskellTypeError {
+        final Apply apply = new Apply(new Ident("id"), new Value(this.integer, "42"));
+
+        assertEquals("(id 42)", apply.toHaskell());
+        assertEquals(this.integer.toHaskellType(), apply.analyze(this.env, this.genSet).prune().toHaskellType());
```
wandernauta: A helper method `assertHaskellTypeEquals` that does the relevant analyzing, pruning and toHaskellTypeification might be neat.

Code/src/test/java/nl/utwente/group10/haskell/expr/ApplyTest.java
```diff
@@ -1,64 +1,95 @@
...
-//        assertEquals(a2.getType(), T_string);
-//        assertEquals(a2.toHaskell(), "massive 1.2 0 \"a\" (4.5, 3, \"b\")");
-//
-//        // Third application
-//        Apply a3 = new Apply(massive, new Value(T_float, "1.2"), new Value(T_integer, "0"),
-//                new Value(T_string, "\"a\""), new Value(new TupleT(T_a, T_b, T_c), "(4.5, 3, \"b\")"));
-//        assertEquals(a3.getType(), a2.getType());
-//        assertEquals(a3.toHaskell(), a2.toHaskell());
-//    }
+    @Test(expected=HaskellTypeError.class)
```
wandernauta: Might want to split this test in two: as it stands, the last assert is probably never tried.

Code/src/test/java/nl/utwente/group10/haskell/expr/ExprTest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    public final void testAnalyze() throws HaskellTypeError {
+        assertEquals("[(Int -> Int)]", this.expr.analyze(this.env, this.genSet).prune().toHaskellType());
+    }
+
+    @Test
+    public final void testToHaskell() {
+        assertEquals("((map (*)) [1, 2, 3, 5, 7])", this.expr.toHaskell());
```
wandernauta: This might not be the best example: `(*)` takes two arguments, which means that the above will not work.

Code/src/test/java/nl/utwente/group10/haskell/expr/IdentTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    public final void testToHaskell() throws HaskellTypeError {
+        assertEquals("id", new Ident("id").toHaskell());
+    }
+
+    @Test(expected=HaskellTypeError.class)
+    public final void testIncorrectName() throws HaskellTypeError {
+        new Ident("id").analyze(new Env(), new GenSet());
+        assertTrue("Invalid state reached: exception not raised.", false);
```
wandernauta: The `assertTrue` should not be necessary: the `expected` in Test makes the test pass *if and only if* the exception is thrown.

Code/src/test/java/nl/utwente/group10/haskell/type/ConstTTest.java
```diff
@@ -0,0 +1,24 @@
+package nl.utwente.group10.haskell.type;
+
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+public class ConstTTest {
+    @Test
+    public final void testToHaskellType() {
+        final ConstT integer = new ConstT("Integer");
+        final ConstT integerWithArg = new ConstT("Integer", integer);
```
wandernauta: `Maybe Int` might be a good example, i.e. what `lookup 10 [(Int, Int)]` gives you.

wandernauta:
This! I like this. Tested, documented, decent code throughout. Also, more deletions than additions. [Much approved.](https://i.imgur.com/NxDUeUB.gif)

The above comments are almost all small doc-related nags. I'll merge with or without those fixed, your call.

Thanks! :sparkles:

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
jjkester: Nothing is evaluated. Until you call analyze(), we don't even know if the application is valid. I would call that pretty lazy. So either ```Apply``` is lazy, or it should get its ass kicked.

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
jjkester: If he/she feels like it. I intended it as a warning - one might want to chain multiple applies first and then call ```analyze``` (or just ignore to do it). Their choice, however, users of any program implementing this library may not be happy with that decision.

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
jjkester: It is equal only in this case, in other ```Expr``` classes there is a difference between the output of the two methods. I chose to keep this separate in case we might change either of them and not realize that the other method's output changes as well.

Code/src/main/java/nl/utwente/group10/haskell/expr/Expr.java
jjkester: Also, the ```GenSet``` class is _not_ documented at all...

Code/src/main/java/nl/utwente/group10/haskell/expr/Expr.java
jjkester: If I understand Java correctly, this means that within the scope of ```analyze``` the variable ```genSet``` cannot be assigned another object. Seems logical at this point, we can change this later if needed (more final is more better).

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
jjkester: Fixed, wrote some documentation for the constructor as well (was missing).

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
jjkester: Yes, because it is an identifier. No, because it extends ```Expr```, so it is an expression. I think you should see this as a valid expression as it functions that way. The reference in this case is ```Ident.name```.

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
jjkester: It is equal only in this case, in other Expr classes there is a difference between the output of the two methods. I chose to keep this separate in case we might change either of them and not realize that the other method's output changes as well.

Code/src/main/java/nl/utwente/group10/haskell/hindley/HindleyMilner.java
jjkester: TODO noted. Seems like a solid idea.

Code/src/main/java/nl/utwente/group10/haskell/hindley/HindleyMilner.java
jjkester: I actually quite like them as in these small cases I find it clearer to understand because Java has massive amounts of all kinds of symbols that are required to simply make something conditional.

Just for fun:

```
String name;

if (this.field1 = some_value && this.field2 = this.field1 || this.field2 == null) {
    name = this.name;
} else {
    name = "Unknown";
}

# The same thing:
String name = this.field1 = some_value && this.field2 = this.field1 || this.field2 == null
        ? this.name
        : "Unknown";
```

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
jjkester: Do you have examples?

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
jjkester: Not much, again, I would like to keep the methods separated. Although I begin to realize that my earlier argument does not really hold up... Still, the ```toString``` is not _always_ equal to the ```toHaskell```/```toHaskellType``` methods so I find it much cleaner to not have them depend on one another.

Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
jjkester: True. I'll review it.

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
wandernauta: Hmm, but analyze() doesn't evaluate the expression either, does it? It seems the laziness you describe here seems to be a bit different from Haskell's special kind of laziness, I think that's part of the confusion.

I'd just call it an application.

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
wandernauta: Understood.

Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
jjkester: I'll look into it.

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
wandernauta: Hmm, okay.

Code/src/main/java/nl/utwente/group10/haskell/expr/Expr.java
wandernauta: Sounds reasonable.

Code/src/test/java/nl/utwente/group10/haskell/expr/ApplyTest.java
jjkester: Fixed.

Code/src/main/java/nl/utwente/group10/haskell/hindley/HindleyMilner.java
wandernauta: Yep, agreed.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
wandernauta: "Int" is a type with a nullary type constructor; it has zero arguments.

Code/src/test/java/nl/utwente/group10/haskell/expr/ApplyTest.java
jjkester: The last assert is there in case something goes horribly wrong - it is _probably_ never tried. The  tested aspect is also tested elsewhere.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
wandernauta: In this specific case, you're implementing the same idea twice with entirely different approaches. If you don't want the methods to depend on each other, at least pick one and copy that :)

Code/src/test/java/nl/utwente/group10/haskell/expr/ExprTest.java
jjkester: It would work: you would just end up with a list of ```Int -> Int``` functions. As seen in the ```testAnalyze``` method.

jjkester:
Improved a whole lot of stuff - and learned to never trust an IDE that has just booted up (as per commits ```233ada0a``` and ```e72c5ec```).

Code/src/test/java/nl/utwente/group10/haskell/expr/ExprTest.java
wandernauta: You're absolutely right. A list of functions! Cute.

Code/src/main/java/nl/utwente/group10/haskell/expr/Ident.java
wandernauta: Gotcha.

wandernauta:
Do you want to pick up the few remaining refactoring things as well, or can I go ahead and merge?

jjkester:
I'm fine with the less important things being merged into master.

wandernauta:
Merged! Thanks!

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
jjkester: Ah, yes, of course. Must have still been asleep.

PR #26, opened 2015-03-26
Demonstrate JavaFX data binding

wandernauta:
Hello hello,

I noticed some hardcore manual setting of properties going on in frontend stuff so far. This is not a problem in and out of itself, however, JavaFX also has some data binding features which could be of use. Specifically:

- JavaFX can inject parts of the interface for you by `fx:id`. This replaces `lookup` and friends.
- JavaFX can observe (!) properties from your class and use those as attributes. This replaces `setText` et al.

This has a few advantages:

- No duplicate state: for example, the function name is now stored in one place, not copied from the FunctionBlock to the corresponding label.
- Decoupling: want to add another label with the same content? Easy.
- No need to remember doing `setText()` calls or anything.
- The FXML is a bit simpler and shorter.
- No need to do calls to `this.lookup`. These calls need an (ugly!) cast to do what you want, and they're probably slow besides, as they have to walk through the entire component every time you do a lookup. Injecting happens once.

This PR is just a (runnable) demonstration/proof of concept; I'll close it right away. It is provided in the hope that it will be useful.

Cheers,
Wander

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/Block.java
@@ -1,12 +1,16 @@
 package nl.utwente.group10.ui.components;
 
+import javafx.fxml.Initializable;
 import javafx.scene.layout.StackPane;
+
+import java.net.URL;
+import java.util.ResourceBundle;
 
 /**
  * Base UI Component that other visual elements will extend from.
  * If common functionality is found it should be refactored to here.
  */
-public class Block extends StackPane {
+public class Block extends StackPane implements Initializable {
 
 	/** Selected state of this Block*/
 	private boolean isSelected = false;
@@ -18,5 +22,9 @@ public class Block extends StackPane {
 	public void setSelected(boolean selectedState) {
 		//TODO If another object is selected then deselect it first!!
 		isSelected = selectedState;
+	}
+
+	@Override
+	public void initialize(URL location, ResourceBundle resources) {
 	}
 }
--- a/Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
@@ -1,17 +1,13 @@
 package nl.utwente.group10.ui.components;
 
-import java.io.IOException;
-
-import nl.utwente.ewi.caes.tactilefx.fxml.TactileBuilderFactory;
-import nl.utwente.group10.ui.Main;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+import javafx.fxml.FXML;
 import javafx.fxml.FXMLLoader;
 import javafx.scene.Node;
-import javafx.scene.control.Label;
 import javafx.scene.layout.Pane;
-import javafx.scene.layout.Region;
-import javafx.scene.layout.StackPane;
-import javafx.scene.control.Label;
-import javafx.scene.shape.Rectangle;
+
+import java.io.IOException;
 
 /**
  * Main building block for the visual interface, this class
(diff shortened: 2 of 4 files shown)


wandernauta:
@TheMSB Great!

On a general note, please don't merge this branch - it breaks tests both locally and on Travis.

Code/src/main/java/nl/utwente/group10/ui/components/Block.java
```diff
@@ -1,12 +1,16 @@
 package nl.utwente.group10.ui.components;
```
wandernauta: I'm not sure what `Initializable` is, but the injector seems to need it.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -21,34 +17,31 @@
...
-	 * @return a new instance of this class
-	 * @throws IOException
-	 */
-	public static FunctionBlock newInstance(int numberOfArguments) throws IOException {
-		FunctionBlock functionBlock = (FunctionBlock) FXMLLoader.load(Main.class.getResource("/ui/FunctionBlock.fxml"), null, new TactileBuilderFactory());
-		functionBlock.initializeArguments(numberOfArguments);
 
-		return functionBlock;
+	/** The name of this Function. **/
+	private StringProperty name;
```
wandernauta: Note that this is a `StringProperty`, not a string.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -21,34 +17,31 @@
...
+	/** The type of this Function. **/
+	private StringProperty type;
+
+	@FXML
+	private Pane nestSpace;
+
+	public FunctionBlock(int numArgs) throws IOException {
+		FXMLLoader fxmlLoader = new FXMLLoader(getClass().getResource("/ui/FunctionBlock.fxml"));
+		fxmlLoader.setRoot(this);
+		fxmlLoader.setController(this);
```
wandernauta: This is where the magic happens.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -21,34 +17,31 @@
...
 
-		return functionBlock;
+	/** The name of this Function. **/
+	private StringProperty name;
+
+	/** The type of this Function. **/
+	private StringProperty type;
+
+	@FXML
+	private Pane nestSpace;
```
wandernauta: nestSpace will automatically be filled from whatever has a `fx:id` of nestSpace.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -21,34 +17,31 @@
...
-	 * @return a new instance of this class
-	 * @throws IOException
-	 */
-	public static FunctionBlock newInstance(int numberOfArguments, String name) throws IOException {
-		FunctionBlock functionBlock = newInstance(numberOfArguments);
-		functionBlock.setName(name);
-		
-		return functionBlock;
+
+	public static FunctionBlock newInstance(int numArgs) throws IOException {
```
wandernauta: Not really necessary, included for compatibility to keep the diff small.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -64,8 +57,7 @@ public String executeMethod() {
 	 * @param node to nest
 	 */
 	public void nest(Node node) {
-		Pane nestSpace = (Pane) this.lookup("#nest_space");
-		((Label) this.lookup("#label_function_name")).setText("Higher order function");
+		name.set("Higher order function");
```
wandernauta: This will set the name property to the given string.

Behind the scenes, the actual visible label is bound to (set to observe) this property. It will automatically update itself to show the new value, if needed.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -64,8 +57,7 @@ public String executeMethod() {
...
 		nestSpace.getChildren().add(node);
```
wandernauta: Look ma, no `lookup`!

Code/src/main/resources/ui/FunctionBlock.fxml
```diff
@@ -1,28 +1,20 @@
...
-<FunctionBlock xmlns:fx="http://javafx.com/fxml/1" styleClass="block">
-	<BorderPane id="foreground" minHeight="90" minWidth="180">
+<fx:root type="nl.utwente.group10.ui.components.FunctionBlock" xmlns:fx="http://javafx.com/fxml/1" styleClass="function, block">
+	<BorderPane minHeight="90" minWidth="180">
 		<top>
 			<BorderPane>
 				<top>
-					<Label id="label_function_name" text="Function name"
-						styleClass="functionBlockTitle" />
+					<Label text="${controller.name}" styleClass="title" />
```
wandernauta: Notice the `"${controller.name}"` binding expression. Also notice that the label does not have an id.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -81,20 +73,24 @@ private void initializeArguments(int numberOfArguments) {
...
-		label.setText(functionName);
+		this.name.set(name);
+	}
+
+	public String getType() {
+		return type.get();
+	}
+
+	public void setType(String type) {
+		this.type.set(type);
```
wandernauta: Getters are mandatory.

Code/src/main/java/nl/utwente/group10/ui/components/FunctionBlock.java
```diff
@@ -1,17 +1,13 @@
...
-import java.io.IOException;
-
-import nl.utwente.ewi.caes.tactilefx.fxml.TactileBuilderFactory;
-import nl.utwente.group10.ui.Main;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+import javafx.fxml.FXML;
 import javafx.fxml.FXMLLoader;
 import javafx.scene.Node;
-import javafx.scene.control.Label;
```
wandernauta: Note that we no longer need to import Label: FunctionBlock is now blissfully unaware of the concept of labels.

PR #28, opened 2015-03-27
Refactored FXML, JUnitTestFixes, Dynamic Labeling

TheMSB:
Refactored the FXML as suggested, all Block Extension classes now properly support the new format.
JUnitTests have been tailored accordingly and now work with the new code.
Dynamic Labeling that is now possible because of new FXML format has been implemented in DisplayBlock and ValueBlock.


wandernauta:
Looks reasonable :+1:  (It also looks quite a lot like the proof-of-concept, but whatever.) A few things:

- Drop the newInstance methods, as they make it confusing whether one should call those or the constructor.
- Changing the background color of function blocks, was that an intentional change? I kinda liked the alice blue.
- This approach seems to (still) assume a push-based model, where blocks push their results into their outputs. I have to ask you: are you *very* certain that's the most sensible way to go about it?
- Quite a bit of copy/paste-action going on in those constructors; you could pull some of that up into the constructor for Block, I think.
- Property getters are missing. While the code seems to work without them, I'm not sure how.

TheMSB:
All but the Push/Pull issues have been addressed. 
Seems like a good idea that the Push/Pull idea will be discussed on Monday during general meeting.

wandernauta:
Looks good, thanks! :sparkles:

PR #29, opened 2015-03-28
Add TypeCheckerChecker

wandernauta:
The TypeCheckerChecker is a simple GUI interface to the 'type inference' (major
air quotes there) part of the backend, which allows fiddling around with it for
a bit to see if it makes any obvious mistakes.

Screenshot


wandernauta:
Test failure is unrelated.

jjkester:
Looks solid enough, changes to typeparser are useful, so I would not be disappointed if this is merged.

(The type checker checker could maybe be/become a "tool" in the application? ;-) )

PR #46, opened 2015-04-14
MainMenu modification for Value and Display Blocks

TheMSB:
This is a proof of concept to modify the MainMenu to include Value and Display Blocks. Currently I am unsatisfied with the code. Please review and advise on possible refactoring and changes.


wandernauta:
I don't really see anything wrong with this. What's causing your lack of satisfaction?

wandernauta:
You could move the `try`/`catch` from the `add*Block` methods to the `addBlock` method, but not matter where you place it, those are all just waiting for a proper FXML loader so that constructors don't `throw` IOError anymore.

TheMSB:
I feel like the implementation I wrote is not future proof, I will process your feedback on the try/catch displacement. If you feel it is adequate enough then feel free the merge once I have committed.

PR #47, opened 2015-04-14
UI package refactoring and whitespace fixing

TheMSB:
Created a new separate package for Block components since their number was growing to be quite large. Fixed whitespace for all of the UI classes that were using tabs.


wandernauta:
Looks great! Will merge when Travis is green.

Thanks! :+1:

TheMSB:
No problem, thanks for the review.

PR #50, opened 2015-04-17
Type class support

jjkester:
Basic, 'Fake' support of type classes.

VarTs have type classes as constraints. Combining two VarTs means the intersection of their constraints. Substitution of a VarT by a ConstT means that the ConstT must be in the type classes of the VarT.


Code/src/main/java/nl/utwente/group10/haskell/expr/Expr.java
```diff
@@ -26,6 +26,19 @@
...
 
     /**
+     * Analyzes the type tree and resolves the type for this usage of this expression, using an empty GenSet. This
+     * method can be used to call when analyzing the root of the expression tree (the first step in the type inference
+     * of an expression).
+     *
+     * @param env The current Haskell environment.
+     * @return The type for this usage of this expression.
+     * @throws HaskellException The type tree contains an application of an incompatible type.
+     */
```
wandernauta: It could be an idea to make the other analyze() protected to avoid any confusion.

Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
```diff
@@ -1,12 +1,55 @@
...
+import nl.utwente.group10.haskell.type.TypeClass;
 
 /**
  * Haskell Environment.
  */
-public class Env extends HashMap<String, Type> {
-}
+public class Env {
+    /**
+     * Map containing the types of expressions known to the environment.
```
wandernauta: Terminology: aren't all the expressions known to the environment functions? For `HaskellCatalog`, I've assumed they would be (and put them in `functions.xml`). Maybe a `private Map<String, Type> functions;` would be clearer. Maybe not.

Code/src/main/java/nl/utwente/group10/haskell/type/VarT.java
```diff
@@ -49,6 +60,35 @@ public final String getName() {
...
+            for (TypeClass typeclass : this.constraints) {
+                if (typeclass.hasType(type)) {
+                    out = true;
+                }
+            }
+        } else {
+            out = true;
+        }
+
+        return out;
```
wandernauta: I'd be in favor of dropping the one-return rule here and going (psuedocode 'cause I'm lazy):

````
for (tc : this.c) {
  if (tc.hasType(type)) {
    return true;
  }
}

return false;
````

wandernauta:
This looks very reasonable. I wouldn't say I'm completely certain that it does the right thing in every circumstance, but I like where you're going with this. Would love to play with it once you get it to compile.

Great job so far! :sparkles:

Code/src/main/java/nl/utwente/group10/haskell/expr/Expr.java
jjkester: That would break existing code, so I'll leave it at this.

Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
jjkester: Not all Exprs are contained in the environment (where environment refers to the Env instance), only Idents when properly used.

jjkester:
Basic test added, no longer WIP, but looking for a test proposal to throw at it (suggestion welcome!). I consider this PR complete once a second, more thorough test has been added (and runs without failure).

wandernauta:
Neat! Let me know when you've added the second test.

A nice simple test would be `succ`: `succ (10 :: Int)` should unify to Int, not Num, and `succ "hello"` shouldn't unify at all. (A terribly complicated test would be `truncate` or `(<*>)` but I think we just want to avoid those.)

To make this really useful, I think TypeParser would need to be extended somewhat to have it parse the `Num a => a` syntax, and functions.xml would need to be changed to have their real types.

jjkester:
Added second test. Adapting the parser is a bit too much to do today. Is it an idea to keep this PR open until the type class support is complete?

wandernauta:
If it doesn't break anything, I'd be happy to just merge it and do the parser in a new PR.

The second test doesn't show up for some reason, are you sure you've pushed it?

jjkester:
It seems I have ignored the message from my IDE that told me that the push had failed ;-)

wandernauta:
Gemerged, thanks! :shipit:

PR #52, opened 2015-04-17
Report: Add a process chapter, stubs, cover image

wandernauta:
This PR adds some stubs (beginnings of paragraphs, headers...) to the report,
hopefully to inspire people to start writing, plus the beginning of a chapter
on the process we are using. 

It also adds a cute fox as a cover image.


wandernauta:
To see the cute fox in all her snowy glory, check out the [PDF render](https://www.sharelatex.com/github/repos/wandernauta/Groep10).

jjkester:
Any reason for the more than one empty page at the end?

wandernauta:
Are you sure you're not looking at the wrong PDF? ShareLaTeX builds all commits, not only those where the report changes, so the green button on the link above might lead to a different commit. 

Here's how it renders on my machine: http://twnc.org/d/report.pdf

wandernauta:
Never mind, it seems ShareLaTeX builds all commits, then throws some of them away for no reason. The link above still works, though.

I've applied the suggested changes.

wandernauta:
The *vulpes vulpes* covergirl is just a placeholder, really - when the UI has crystallized a bit, we might put a screenshot on the front.

PR #53, opened 2015-04-18
UI refactoring

TheMSB:
Refactored several classes:

 * Each UI Component type now resides in their respective package, Blocks in blocks, Anchors in anchors etc.

 * I noticed that all of the UI Component classes had duplicate code in their constructors, attempted to refactor by creating abstract class Component, sadly I ran into complications with ConnectionLine and Circle because of multiple inheritance, I tried composition but it seems hard to implement with FXML access.

 * Blocks respective toString method has now been refactored into a one serves all toString in Block.

 * Attempted to create a more generic setAnchor method for Connection but am having difficulties because of the Optional Generics.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
@@ -7,9 +7,9 @@ import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.scene.Node;
 import nl.utwente.ewi.caes.tactilefx.control.TactilePane;
-import nl.utwente.group10.ui.components.Connection;
 import nl.utwente.group10.ui.components.blocks.Block;
 import nl.utwente.group10.ui.components.blocks.DisplayBlock;
+import nl.utwente.group10.ui.components.lines.Connection;
 import nl.utwente.group10.ui.handlers.ConnectionCreationManager;
 
 /**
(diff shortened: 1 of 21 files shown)


wandernauta:
Looks good overall. A few things:

- Even with your refactored Component class, constructors still throw IOExceptions. Do you have an idea how we can get rid of those? 
- The new `setAnchor` looks needlessly complex. You shouldn't have to cast Optionals at all. Check out the Java 8 docs if you haven't already - if you have, it might help to think of Optional as an ArrayList that always has one or zero elements.

TheMSB:
This last commit should fix the last few things. Final notes:
 * setAnchor works but has potential for further refactoring.
 * catch clause in base components needs useful catch (or we need to refactor FXMLLoading even further.

I will introduce no further changes but will rebase the commits so only code that was strictly needed ends up in master.

wandernauta:
The ComponentLoader 'interface' doesn't seem to fix the IOException problem. It's also not really a loader: the actual loading only happens when `.load()` is called on the result.

Maybe I'm just not 'getting it', but could you explain what the ComponentLoader does and why it's useful?

TheMSB:
The idea behind ComponentLoader is that it provides a default implementation for the loading of the FXML. Now each Component only has to use one method to instantiate it's FXML instead of having the same 4 lines in each constructor. 

On a side note the reason .load() is called on the result instead of being called right away is to prevent FXML complaining about the root of the FXMLLoader. If you want to see what happens try calling .load() right away.

wandernauta:
Okay.

PR #55, opened 2015-04-21
Type class support for type parser, catalog, ...

jjkester:
Open issues:

- [x] Let UI use the new ```HaskellCatalog``` instead of the old (refactored) ```HaskellFunctionCatalog```
- [x] Clean up ```catalog``` package
- [ ] Add type classes to functions in ```catalog.xml```
- ~~Change ```toHaskellType()``` on ```VarT``` and make it parsable (nice to have)~~


wandernauta:
Thanks! I'll take a look right away.

Code/src/main/antlr4/nl/utwente/group10/haskell/typeparser/Type.g4
```diff
@@ -4,14 +4,20 @@ CT : [A-Z][a-z]+ ;
...
+tupleType : '(' innerType (',' innerType)+ ')' ; // tuple type, k>=2
+listType : '[' innerType ']' ;              // list type
+parenType : '(' innerType ')' ;             // type with parentheses
 
-constantType : typeConstructor (WS* type)* ;
+constantType : typeConstructor (WS* innerType)* ;
 typeConstructor : CT ;
 variableType : VT ;
+
+typeClasses : '(' typeWithClass (',' typeWithClass)* ')' '=>' ;
```
wandernauta: I just found out that if there is only one typeclass constraint, you can also write

    f :: Num a => a
    f = 10

That is, without the parentheses. The typeClasses rule would then become something like

    typeClasses : '(' tWC (',' tWC)+ ')' '=>' | tWC '=>' ;

I think.

Code/src/main/antlr4/nl/utwente/group10/haskell/typeparser/Type.g4
```diff
@@ -4,14 +4,20 @@ CT : [A-Z][a-z]+ ;
...
 
-tupleType : '(' type (',' type)+ ')' ; // tuple type, k>=2
-listType : '[' type ']' ;              // list type
-parenType : '(' type ')' ;             // type with parentheses
+tupleType : '(' innerType (',' innerType)+ ')' ; // tuple type, k>=2
+listType : '[' innerType ']' ;              // list type
+parenType : '(' innerType ')' ;             // type with parentheses
 
-constantType : typeConstructor (WS* type)* ;
+constantType : typeConstructor (WS* innerType)* ;
```
wandernauta: Are we sure this `WS*` is doing anything?

Code/src/main/java/nl/utwente/group10/haskell/catalog/ClassEntry.java
```diff
@@ -0,0 +1,70 @@
...
+import nl.utwente.group10.haskell.type.Type;
+import nl.utwente.group10.haskell.type.TypeClass;
+import nl.utwente.group10.haskell.typeparser.TypeBuilder;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/** A class entry in the Haskell catalog. */
+public class ClassEntry implements Comparable<ClassEntry> {
```
wandernauta: Could perhaps use a common superclass `CatalogEntry` - they're both named *somethings* that get sorted on their names.

Code/src/main/java/nl/utwente/group10/haskell/catalog/HaskellClassCatalog.java
```diff
@@ -0,0 +1,132 @@
...
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.validation.SchemaFactory;
+import java.io.IOException;
+import java.net.URL;
+import java.util.*;
+
+/**
+ * Provides a convenient interface to the XML catalog of available Haskell classes.
+ */
+public class HaskellClassCatalog {
```
wandernauta: I'm not sure how I feel about splitting the Catalog in two - it seems you'd have to duplicate quite some XML-related nonsense in doing so. A possible alternative would be to have one catalog that knows about both functions and typeclasses, a `catalog.xml` if you will:

````
<catalog>
  <functions>
    <!-- as before -->
  </functions>
  <typeclasses>
    <!-- as here -->
  </typeclasses>
</catalog>
````

File: Code/src/main/java/nl/utwente/group10/haskell/env/Env.java [hunk 5]
```diff
@@ -23,9 +23,17 @@
...
+     */
+    public Env(Map<String, Type> exprTypes, HashMultimap<Type, TypeClass> typeClasses) {
+        this.exprTypes = exprTypes;
+        this.typeClasses = typeClasses;
+    }
+
     public Env() {
-        this.exprTypes = new HashMap<String, Type>();
-        this.typeClasses = HashMultimap.create();
+        this(new HashMap<String, Type>(), HashMultimap.create());
```
wandernauta: An empty `Env` is not all that useful - would it make sense to make the constructors for Env private?

Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
```diff
@@ -52,4 +60,28 @@ public final void addTypeClass(TypeClass typeclass) {
...
+    }
+
+    /**
+     * Joins two environments in a special way. Returns a new Env with the expression types from the first Env and the
+     * type classes from the second Env.
+     * @param exprEnv The first environment.
+     * @param classEnv The second environment.
+     * @return A new environment containing the expressions from the first Env and the type classes from the second Env.
+     */
+    public static Env join(Env exprEnv, Env classEnv) {
```
wandernauta: This would be a side effect of splitting the catalog, right?

Code/src/main/java/nl/utwente/group10/haskell/typeparser/TypeBuilder.java
```diff
@@ -2,15 +2,33 @@
...
+
+    /**
+     * @param typeClasses The available type classes.
+     */
+    public TypeBuilder(Map<String, TypeClass> typeClasses) {
+        this.typeClasses = Optional.ofNullable(typeClasses);
+    }
+
+    public TypeBuilder() {
+        this(null);
```
wandernauta: If you do the `new HashMap<>` here instead of in TypeBuilderListener, you could get rid of the `Optional` and make the next hunk a bit simpler.

wandernauta:
Looks pretty sensible! Only thing I'm on the fence about is the catalog split, but other than that it looks great.

Thanks a lot! :star2:

Code/src/main/antlr4/nl/utwente/group10/haskell/typeparser/Type.g4
jjkester: I'll add the possibility.

Code/src/main/antlr4/nl/utwente/group10/haskell/typeparser/Type.g4
jjkester: Not sure, but it was already there. I like it there because as reader you know there *should* be whitespace in the input.

Code/src/main/java/nl/utwente/group10/haskell/catalog/ClassEntry.java
jjkester: I was planning on doing that with the cleanup of the package.

Code/src/main/java/nl/utwente/group10/haskell/catalog/HaskellClassCatalog.java
jjkester: I have thought about that too, but for know I was looking for a quick-'n-dirty solution to get it working quickly. (In fact, I started working on this but the existing ```HaskellCatalog``` class needed to much adaptation, so I chose to just copy-paste it.) Also, this would break functionality for the UI, so this has to be done very carefully and introduces a lot of work (which I am happy to do, but again, a quick solution for now).

This is another part of the cleanup.

Code/src/main/antlr4/nl/utwente/group10/haskell/typeparser/Type.g4
wandernauta: That's true. On the other hand, it could give the impression that extra whitespace is *not* allowed elsewhere, which it is.

Code/src/main/antlr4/nl/utwente/group10/haskell/typeparser/Type.g4
wandernauta: Great!

Code/src/main/java/nl/utwente/group10/haskell/catalog/ClassEntry.java
wandernauta: Understood.

Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
jjkester: No. The constructors of Env are used in other packages (catalog for example) and tests where an empty Env is needed.

Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
jjkester: Yes, completely obsolete once the cleanup of the catalog is done.

Code/src/main/java/nl/utwente/group10/haskell/typeparser/TypeBuilder.java
jjkester: Will look into it.

wandernauta:
Can you give me a heads-up when you've done the cleanup? Muchísimas gracias.

jjkester:
Clean-up finished.

Sorry for the many merge commits - while working on the code I realized I needed the last UI state.

Changing VarT to make its string type representation parsable is quite difficult without breaking other things, so I'll leave that.

jjkester:
No longer WIP, please check if it meets all quality standards :-)

Thank you!

Code/src/main/resources/catalog/catalog.xml
```diff
@@ -0,0 +1,68 @@
...
+    <classes>
+        <class name="Num">
+            <instance name="Int"/>
+            <instance name="Float"/>
+            <instance name="Double"/>
+        </class>
+    </classes>
+    <functions>
+        <category name="Math">
+            <function name="(*)" signature="Float -> Float -> Float"></function>
```
wandernauta: These should probably be `Num a => a -> a -> a` now.

wandernauta:
This looks fantastic! I love the new style for the Catalog. Two minor niggles (and they really are minor):

- Sorting functions in the menu seems a bit wonky now.
- The functions in the catalog still pretend to use Floats only.

Should be easy to fix, will merge posthaste and with great gratitude with those changes. Verily.

Code/src/main/resources/catalog/catalog.xml
jjkester: I was too quick - the functions in the catalog still need to use type classes.

jjkester:
Fixed :-)

Is it correct that all higher order functions have been removed?

wandernauta:
Merged! Pleasure doing business.

As for the highter order functions: I more or less picked a random selection of simple functions for the calculator demo and copied those by hand - so it's not so much that higher order functions have been removed as that I didn't bother to copy those over.

PR #57, opened 2015-04-22
Fixes for type classes

jjkester:
Fixes the issues where...

- ... a typeclass contains an uppercase character after the first position (e.g. ```RealFrac```). This was an issue with the lexer.
- ... an unknown typeclass is ignored by the parser where it unintentionally falled back to the last used type class.


wandernauta:
Looks good, will merge on green.

PR #58, opened 2015-04-28
Intersecting Error Lines

TheMSB:
* Created proof of concept for drawing intersecting error lines for Connections.

Not to be merged yet.


TheMSB:
@wandernauta and @i3anaan, after the meeting today it seems that this exact implementation is not wanted if I'm not mistaken. I will leave the branch as is in the event this functionality is desired. A more general TypeMismatch branch will be made.

wandernauta:
Thanks! I'll close it for now, we'll know where to look if we ever want to do something like this.

PR #59, opened 2015-04-28
TypeMismatch indications for Connections

TheMSB:
Right now connection lines will turn Orange for their error state, since there is no clear implementation yet to provide TypeMismatch errors the visualisation for specific errors has not yet been implemented.


Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
```diff
@@ -42,6 +42,9 @@ public final void errorAll() {
         for (Node node : getChildren()) {
             if (node instanceof Block) {
                 ((Block)node).error();
+            } else if (node instanceof Connection) {
+                System.out.println("derp");
```
wandernauta: This derp is probably not crucial :)

wandernauta:
Merged (without the derp, that is). Thanks!

PR #61, opened 2015-05-01
Base concept for block context menu

TheMSB:
Bare implementation as of yet but should give a clear idea what to expect.


jjkester:
Be aware that your code is not compiling correctly or that tests are failing. Please look into that.

wandernauta:
Hey there! The new menu looks really pretty :)

Is the idea that this menu will replace the current (temporary?) main menu? Or that it would work alongside it?

Whatever the case may be, here's a few things to keep in mind:

- There's something strange going on with the resource path, it seems. Your file is called `Undo.png`, not `undo.png`, but even when I change that and move the file to the resource root, I can't get it to work reliably in Maven *and* IDEA. I'm going to try and see what's going on here, but if you've got any ideas they are more than welcome.
- Whitespace changes in `Block.java` make it a bit hard to see what's really changed. If you really want to make those whitespace changes, please do so in a seperate commit/PR.
- Both this 'context' menu and the 'real' ContextMenu/main menu appear on right-click. This may not be what you want.
- The menu is drawn with an opaque white background on my machine: [link](https://i.imgur.com/o697kCt.png). Is that by design?

And then some small things:

- I personally don't really think this menu needs its own package. MainMenu doesn't have one either.
- It would probably be convenient if you came up with a more descriptive name - ContextMenu is the name of a class in JavaFX, which we use for the main menu.
- You can make ContextMenu's constructor a bit simpler if you use Java 8's new lambda syntax instead of all that `new EventHandler<ActionEvent>()` drivel.

If you've got any questions, do not hesitate to ask.

TheMSB:
@jjkester  I believe all of the tests should be working, the Travis build error failure seems to be because of the rollback on the Travis patch, all test are running fine locally.

@wandernauta 

 * The idea for this menu is to work alongside the current main menu.
 * The resource path originally gave me some problems too, I put the image file inside the same folder as the java source file (this worked) I then moved it to the more appropriate resource path where it also works. When I was originally moving the file around it seemed that the IDE was giving me problems whenever I did NOT refresh the entire project and cleaned it. It might be redundant advice but it could help.

 * Whitespace changes are noted, I will factor those out and make a separate PR for those.
 * It is indeed not intended functionality for both of them to appear on right click. (originally this was since main menu was supposed to disappear eventually) I will enforce left clicks for the opening of the contextMenu until a substitute for main menu has been made.

 * The white background you are experiencing should not be happening, it should in fact be transparent. I will try and look into why this might be happening.
 * The package: I intend to eventually implement the **corner** style menu I presented earlier, that combined with the implementation of a function drawer style menu would justify a new package. If you disagree or think I should not implement the package as of yet please let me know.
 * I will work out a better class name.
 * The lambda syntax will indeed simplify the constructor, as such I will implement needed changes.

wandernauta:
Great! Looking forward to seeing those changes. 

You could consider throwing the current  MainMenu in the menu package, I think that would make sense.

As for the resource path issue: the most important thing is that it works in Maven, as the Maven-generated jar will be the version we're going to hand in. I looked a bit closer; if you move the file to the same directory as Main.fxml, use the same '/ui/' prefix, and fix the case, it should be fine (and Eclipse should pick it up as well). If you're confused, do a 'mvn package' and open up the resulting naar in a program that reads zip files, it'll show you what goes where and why the '/ui/' is important in our current setup.

wandernauta:
Correction: 'naar' should of course be 'jar'.

TheMSB:
I'd like some advice on the resource pathing then.

My current implementation uses: 
`this.getClass().getResourceAsStream("undo.png"))));`

Current implementation in Main:

`this.getClass().getResource("/ui/Main.fxml")`

My problem: When using option 2 I get errored with the message that the contractor for Image is not visible, when using option 2 like so:

`new Image(this.getClass().getResourceAsStream("/ui/undo.png"))));`

it also doesn't seem to work, contrast to expectations after reading the java doc of getResourceAsStream: 

`Before delegation, an absolute resource name is constructed from the given resource name using this algorithm:

If the name begins with a '/' ('\u002f'), then the absolute name of the resource is the portion of the name following the '/'.
Otherwise, the absolute name is of the following form:
modified_package_name/name
Where the modified_package_name is the package name of this object with '/' substituted for '.' ('\u002e').`

Any advice as to how to go about this?

wandernauta:
Did you move undo.png to the same folder as Main.fxml yet?

TheMSB:
I most certainly did, to avoid any issues with eclipse I refreshed and cleaned the project twice.

wandernauta:
And you did a Maven run as well? 'mvn process-resources' or the Eclipse equivalent?

TheMSB:
Yes

wandernauta:
If you push what you have now I'll take a look first thing tomorrow.

TheMSB:
I solved the issue, shamefully overlooked a line.

wandernauta:
Great! Let me know when you want me to take another look.

TheMSB:
@wandernauta I tidied up the code a bit and I feel like most of the code for the contextMenu is done.
Some points however:

 * I am waiting on @Kristel-92 for menu icons to tidy up the appearance.
 * Features I feel like they should be in the menu:
    * Copy
    * Paste
    * Delete
    * Save

Copy and Paste would require the implementation of a clipboard and the ability to clone objects. Thinking about the implementation I came up with the following questions: (1) Do we just want to copy the single block or do we want to copy the entire tree above the selected block? (2) A good way to implement cloning for blocks seems to be a way to create new block(s) from an asExpr string, do you agree? (3) I feel like copy/paste functionality is non-essential as of yet and as such I feel like implementation of these functions should be postponed until a later PR. Would you agree or do you believe the implementation to be trivial enough to implement within this PR and sprint?

I feel like the Save functionality comes down to the same thing as the Copy/Paste implementations only then with the addition that we store said object in a catalog of sorts.

wandernauta:
I'll take a look, one sec.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/Block.java
```diff
@@ -2,29 +2,35 @@
...
 public abstract class Block extends StackPane implements ComponentLoader {
 
-    /** The output of this Block.**/
+    /** The output of this Block. **/
     private OutputAnchor output;
-
     /** The pane that is used to hold state and place all components on. */
     private CustomUIPane parentPane;
+    /** The context menu associated with this block instance. */
+    private CircleMenu contextMenu;
```
wandernauta: I'd call it a circle menu consistently to, again, avoid confusion with the main menu which is a ContextMenu.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/Block.java
```diff
@@ -32,41 +38,55 @@ public Block(CustomUIPane pane) {
...
+        this.addEventHandler(MouseEvent.MOUSE_CLICKED, this::handleMouseEvent);
 
-        this.addEventHandler(MouseEvent.MOUSE_CLICKED, this::select);
+        this.createContextMenu();
     }
 
-    /** Sets this block as the selected block. */
-    private void select(MouseEvent mouseEvent) {
-        parentPane.setSelectedBlock(this);
+    private void createContextMenu() {
```
wandernauta: You could probably inline this method.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/Block.java
```diff
@@ -32,41 +38,55 @@ public Block(CustomUIPane pane) {
...
 
-    /** @return the parent pane of this Component. */
+    /**
+     * Sets this block as the selected block. When this block has already been
+     * selected it spawns a contextMenu instead.
+     */
+    private void handleMouseEvent(MouseEvent t) {
+        if (parentPane.getSelectedBlock().isPresent()
+                && parentPane.getSelectedBlock().get().equals(this)
+                && t.getButton().equals(MouseButton.PRIMARY)) {
```
wandernauta: Will this work with touch?

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
```diff
@@ -0,0 +1,99 @@
...
+import jfxtras.scene.layout.CircularPane;
+import jfxtras.scene.menu.CirclePopupMenu;
+
+/**
+ * Circle menu is a context based menu implementation for {@link Block} classes.
+ * Preferably each block class has it's own instance of circle menu. When a
+ * block based class has significant differences to other block classes that
+ * result in different context based actions it should use a specialized
+ * extension of circle menu instead of this one.
+ * <p>
```
wandernauta: Random unclosed `<p>` tag?

Code/src/main/java/nl/utwente/group10/ui/menu/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Contains components for the menus that can be used to visually program a Haskell application.
+ */
+package nl.utwente.group10.ui.menu;
```
wandernauta: Missing newline EOF.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/Block.java
TheMSB: If you mean the createContextMenu then I left it out on purpose, the idea is that block implementations with unique features can override the method to call their own specific CircleMenu implementation. (I imagine stuff like ValueBlock being able to adjust it's values etc.)

wandernauta: Then you probably want it to be `protected`, not `private`, right?

Code/src/main/java/nl/utwente/group10/ui/components/blocks/Block.java
TheMSB: Done.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/Block.java
TheMSB: Noted and changed.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/Block.java
TheMSB: I have no reason to believe it will not since mouse events are cased into touch events. I will however need to test this later this week.

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
TheMSB: Missed, and fixed it.

Code/src/main/java/nl/utwente/group10/ui/menu/package-info.java
TheMSB: added.

wandernauta:
This keeps looking better! To take a stab at answering your questions:

- I think copy and paste would be more useful if there was a feature to select multiple objects/blocks. I'm just not sure how we'd implement that as we're using drag for pan.
- Creating blocks from an asExpr string, that is, a fragment of Haskell code, would require implementing a Haskell parser, right? I'm not sure if that's the easiest way to go about it. I think it would be easier to try to seperate the *meaning* of a FunctionBlock a bit from their *representation*. At the moment, a FunctionBlock is both something visible and something conceptual. Pulling those apart would make stuff like this easier, perhaps.
- Implementing copy/paste would most certainly warrant its own PR.
- I'm not sure I understand the Save feature. What would it do, exactly?

TheMSB:
> I'm not sure I understand the Save feature. What would it do, exactly?

The save feature would store a block into a user specific catalog so that he can reuse his function at a later time in the program (or another one)

wandernauta:
Reuse as in copy? Or as in reference?

Would you save/stash a whole function definition, like the white squares in my mockup or the define blocks in JJ's sketch, or just a random fragment of the tree?

TheMSB:
> Would you save/stash a whole function definition, like the white squares in my mockup or the define blocks in JJ's sketch, or just a random fragment of the tree?

This is exactly the issue I was pondering about, I'm fully aware it won't be implemented now but it was more like food for thought that we can discuss at the next meeting.

wandernauta:
I was thinking more of a global save feature, where you save the entire program before going home and continue working on it the next day, for example. This would probably be in the global menu, however.

TheMSB:
> I was thinking more of a global save feature, where you save the entire program before going home and continue working on it the next day, for example. This would probably be in the global menu, however.

True, but I believe there might be merit in the ability to save parts of your program that you might want to recycle. Maybe you made an awesome function that dissects an entire document and returns only nouns. The next day you realise that maybe you want a separate dissection that gives you a list of all the verbs too. Both functions would essentially have very little difference I am assuming and smaller parts of them could be useful for a range of different applications.

wandernauta:
That does sound kinda useful. I think this would be interesting to discuss at the meeting.

TheMSB:
> That does sound kinda useful. I think this would be interesting to discuss at the meeting.

I believed so too. If there are no other remarks to the code then I will await the graphics from @Kristel-92  and after commit those changes perform a rebase.

wandernauta:
Sounds good!

By the way, the white-background problem was an issue on my end. Transparency started working when I launched a compositor.

TheMSB:
By the way, could you confirm that it is indeed the Travis rollback that caused the test errors? I've tried looking around but haven't made any significant changes that could warrant these errors I believe.

wandernauta:
The problem seems to be this:

> java.lang.IllegalStateException: Not on FX application thread; currentThread = main

This is coming from the jfxtras library. It seems CirclePopupMenu's constructor should be called on the FX thread, not the main thread. In other words, it needs some `Platform.runLater` magic fairy dust (see [here](https://docs.oracle.com/javase/8/javafx/api/javafx/application/Platform.html#runLater-java.lang.Runnable-)).

TheMSB:
> In other words, it needs some Platform.runLater magic fairy dust

I modified the Block constructor to spawn the circle menu using `Platform.runLater`

wandernauta:
:+1: from me. Will merge when the icons arrive.

wandernauta:
Great! I see you've rebased your branch, fantastic.

If I can be entirely honest, I'm unsure about the icons. I had some difficulty seeing/recognizing the icons at first glance. The copy icon is a photocopier, I think? Or a clap board, like at a movie set? The scissors are pretty obvious, but I can't really tell what the last one is supposed to be. This could just be me, however, and I'll merge as-is if you want.

contrast

TheMSB:
> I had some difficulty seeing/recognizing the icons at first glance.

I agree about the icons, we will have to bring up the icon designs with @Kristel-92  though as she is the one that designed them. I just implemented the code, color contrast might not already help a bit to make them more readable.

In short: I prefer to fix the graphics aspect in a separate PR

wandernauta:
Understood. Thanks!

PR #62, opened 2015-05-02
Improvements to type system

jjkester:
- Types have now correct ```equals``` and ```compareTo``` methods
- Ident expressions now use fresh types

Would like input on implementation (improvements, how to clean up code)

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
@@ -1,9 +1,6 @@
 package nl.utwente.group10.haskell.env;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
@@ -60,22 +57,14 @@ public class Env {
 
     /**
      * @param name The name of the expression.
-     * @return The type of the expression if known to the environment, else {@code null}.
+     * @return An Optional containing the Type for the given expression, if any.
      */
-    public final Type getExprType(String name) {
-        return this.exprTypes.get(name);
-    }
-
-    /**
-     * @param name The name of the expression.
-     * @return A new Type object for the type of the expression if known to the environment, else {@code null}.
-     */
-    public final Type getFreshExprType(String name) {
+    public final Optional<Type> getFreshExprType(String name) {
         if (this.exprStringTypes.containsKey(name)) {
             TypeBuilder builder = new TypeBuilder(this.getTypeClasses());
-            return builder.build(this.exprStringTypes.get(name));
+            return Optional.ofNullable(builder.build(this.exprStringTypes.get(name)));
         } else {
-            return null;
+            return Optional.empty();
         }
     }
 
--- a/Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
@@ -7,8 +7,11 @@ import java.util.Set;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
+import nl.utwente.group10.haskell.expr.Expr;
+import nl.utwente.group10.haskell.expr.Ident;
 import nl.utwente.group10.haskell.type.Type;
 import nl.utwente.group10.haskell.type.TypeClass;
+import nl.utwente.group10.haskell.typeparser.TypeBuilder;
 
 /**
  * Haskell Environment.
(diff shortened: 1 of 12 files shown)


Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
```diff
@@ -49,6 +59,41 @@ public Env() {
     }
 
     /**
+     * @param name The name of the expression.
+     * @return The type of the expression if known to the environment, else {@code null}.
```
wandernauta: Would prefer using `Optional<Type>` instead of a hard `null`.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -1,6 +1,7 @@
 package nl.utwente.group10.haskell.type;
 
 import com.google.common.base.Joiner;
+import com.sun.org.apache.xpath.internal.functions.FuncFalse;
```
wandernauta: Is this import required?

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -69,4 +70,26 @@ public String toString() {
             return this.constructor;
         }
     }
+
+    @Override
+    public int compareTo(final Type type) {
+        if (type instanceof ConstT) {
+            if (this.constructor.compareTo(((ConstT) type).constructor) == 0) {
```
wandernauta: For String, you could probably use `equals` instead of `compareTo() == 0`:

> compareTo returns 0 exactly when the equals(Object) method would return true.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -69,4 +70,26 @@ public String toString() {
             return this.constructor;
         }
     }
+
+    @Override
+    public int compareTo(final Type type) {
+        if (type instanceof ConstT) {
```
wandernauta: If you do a quick `ConstT constt = (ConstT) type;` here you can avoid a few casts later.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -69,4 +70,26 @@ public String toString() {
             return this.constructor;
         }
     }
+
+    @Override
+    public int compareTo(final Type type) {
```
wandernauta: Per the Comparable interface, `compareTo` is required to throw a `NullPointerException` if the passed object is `null`. It probably would do just that because you're using `type` later on, but it might be neater to explicitly check for null here.

Code/src/main/java/nl/utwente/group10/haskell/type/VarT.java
```diff
@@ -156,4 +156,9 @@ public final String toString() {
 
         return intersection;
     }
+
+    @Override
+    public int compareTo(final Type type) {
+        return this.equals(type) ? 0 : -1;
```
wandernauta: You're not overriding `equals` as far as I can tell. Does that mean you want `Object.equals`, i.e. VarT's are only equal if they're the same object? That seems reasonable, but could be worthwhile to add as a comment.

wandernauta:
Hey there! This looks well thought out. Two general questions:

- From what I gather, Env now has a `getExprType` which does kinda-sorta the right thing and a `getFreshExprType` which does the right thing all the time but might be slower. I'd prefer to drop `getExprType` and make the slow-but-steady method the only (public) one.
- @i3anaan made a test for (I believe) freshness-related problems [here](https://github.com/wandernauta/Groep10/blob/ui-types/Code/src/test/java/nl/utwente/group10/haskell/type/UnificationTest.java). Does that test pass with these changes?

Good work :+1:

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
jjkester: Not at all, strange that it appears (because I have never used it). Downside of IDE magic I guess.

Code/src/main/java/nl/utwente/group10/haskell/type/VarT.java
jjkester: Yes, that is the expected behaviour. I think the cleanest solution is to override ```equals``` and simply call ```super.equals``` and explain this in the javadoc.

Code/src/main/java/nl/utwente/group10/haskell/env/Env.java
jjkester: That would be nicer, especially because of the ```type != null``` in ```Ident```. Thanks.

Code/src/main/java/nl/utwente/group10/haskell/type/VarT.java
wandernauta: Sounds good.

jjkester:
The test fails on a technicality (expected ```a```, but is a VarT with another letter), but it seems alright.
I have not included the test in the commit (as this would introduce merge conflicts) but I did add ```undefined``` to the catalog.

PR #64, opened 2015-05-03
Implement panning (drag) and zooming (scroll)

wandernauta:
This commit adds panning and zooming functionality.

Pan is implemented by translating the TactilePane in the correct (that
is, opposite) direction. Similarly, zooming is implemented by scaling
the TactilePane.

The code for TactilePane is changed to use pane-relative instead of
scene-relative coordinates at some points, as scene-relative coordinates
do not take into account the transformations applied.

Touch interaction is untested.


jjkester:
Code seems fine :+1: .

I recommend to test this on a touch screen before merging. I am quite sure that at least zooming won't work (but I could be wrong). On the other hand, zooming might be as simple as binding to the right touch gesture/event.

I favor some testing with touch first, but I leave it up to you.

wandernauta:
I'm going to go ahead and merge this, I think being able to pan/zoom with the mouse could be useful during development even if it doesn't work with touch yet. Thanks for taking a look!

PR #68, opened 2015-05-04
Show an alert if a noop evaluation with GHCi fails

wandernauta:
So far, at least four of us have tried running the program on a machine that didn't have GHC installed. This will not work. This PR adds a nice graphical message to the interface itself if there's a problem with ghci.

2015-05-04-101654_676x239_scrot


TheMSB:
Looks great!

PR #69, opened 2015-05-04
Make GhciSession NOT a Singleton

wandernauta:
There were reasons for making GhciSession a singleton, but none of them
were very good. This commit reverts that and has CustomUIPane carry
along an optional GhciSession.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ghcj/GhciSession.java
+++ b/Code/src/main/java/nl/utwente/group10/ghcj/GhciSession.java
@@ -2,7 +2,6 @@ package nl.utwente.group10.ghcj;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.Optional;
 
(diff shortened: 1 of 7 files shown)


TheMSB:
I am still interested in the reasons in favour of a singleton approach in the event you mind sharing them.

wandernauta:
Sure.

Back in the day, during FP class, ghci occasionally took a sweet time (a few seconds) starting up, as it had to load quite a lot of libraries as well as compile some user code.

I wanted to be absolutely sure that the painful multiple-second initialization was only ever done once, not once per test for example. Hence a singleton. However, it turns out that starting ghci with only `Prelude` and `Data.List` imported is pretty fast, and there aren't many tests, so there's really no need.

jjkester:
(Y)

You have my permission to merge this yourself and resolve any conflicts. (If you want me to do it, just let me know.)

wandernauta:
Merged, thanks for taking a look!

PR #73, opened 2015-05-05
Have ComponentLoader swallow/rethrow IOExceptions

wandernauta:
This changes the ComponentLoader interface to itself catch IOExceptions
and rethrow them as RuntimeExceptions.

Not throwing IOExceptions simplifies quite a few methods, as well as the
MainMenu class.

Rationale: When an FXML file can not be read, that's *always* a fatal
error. In other words, the program is useless when it does not have all
its FXML files. In the old implementation, none of the places that
caught the exception did something useful with it.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/components/ComponentLoader.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/ComponentLoader.java
@@ -2,17 +2,22 @@ package nl.utwente.group10.ui.components;
 
 import javafx.fxml.FXMLLoader;
 
+import java.io.IOException;
+
 /**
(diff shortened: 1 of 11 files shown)


TheMSB:
I love the change, it makes  ` ComponentLoader `  much more useful and efficient. Failing to load FXML should indeed be a fatal error since without it you can't really use the UI.

jjkester:
You're totally right. That would indeed clean up a lot of stuff. GJ!

I think it is good practice to document this different approach either with a comment or in Javadoc. Besides that I have no other comments, it is good to merge.

wandernauta:
Done! Thanks for the suggestion. Will merge on green.

PR #75, opened 2015-05-07
Fix contains check in TypeClass (with compareTo)

wandernauta:
Set.contains returns true if its argument is equal to any of the set's
elements, but we do not override equals() or hashCode() on Type. We do
implement Comparable, however, so that is what we use here.

Complexity gone from possible O(1) to definite O(N), but N is small.


i3anaan:
Is dit de fix voor dat probleem wat ik aangekaart heb?

jjkester:
Good eye, this is indeed a problem (and a good solution). GJ! :+1:

PR #76, opened 2015-05-08
Added (and updated) ApplyTest

i3anaan:
The ApplyTest previously in PR 51, also updated (had some bugs).
Due to recent backend fixes the test now succeeds.


wandernauta:
Great! Thanks!

PR #77, opened 2015-05-08
Refactor away CustomUIPane.initialize

wandernauta:
In the current setup, it would be required to call the `initialize`
method on CustomUIPane before doing anything else, i.e. it left the pane
in a limbo state before that method was called. Behaviour when the
method was called multiple times would also be interesting.

This commit replaces the initialize method by adding HaskellCatalog to
the constructor. This means CustomUIPane can no longer be constructed
from FXML, however, that's not much of a loss in our specific case.

It also makes `CustomUIPane.getEnvInstance` a bit simpler and removes
the need for a Main.fxml.


jjkester:
Do you think this will conflict with any changes coming up from for example @i3anaan's branch?

wandernauta:
I'm sure it'll be fine.

jjkester:
Good idea :+1: . It cannot be merged automatically, you can merge it yourself (or ask me to do it).

PR #78, opened 2015-05-08
Minor cleanup: unused imports, unused keywords

wandernauta:


jjkester:
Do you think this will conflict with any changes coming up from for example @i3anaan's branch?

wandernauta:
It might, but I don't think that will be a big problem.

PR #79, opened 2015-05-08
Have FunctionBlock.getOutputType always analyze()

wandernauta:
The current implementation has getOutputType skip analyzing if one or more inputs of the function block are not connected. However, we might not need the disconnected input to say something sensible about the output type. So instead of this:

2015-05-08-184615_638x453_scrot

The PR tells FunctionBlock to always analyze:

2015-05-08-184643_817x528_scrot

As an added bonus, not having to be able to check if inputs are connected simplifies the InputBlock interface a bit.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/components/blocks/DisplayBlock.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/blocks/DisplayBlock.java
@@ -1,9 +1,9 @@
 package nl.utwente.group10.ui.components.blocks;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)


i3anaan:
Dit zit ook in mijn bowtie PR

jjkester:
@i3anaan In that case, this PR seems obsolete. Do you have any idea when to expect your change in master?

wandernauta:
Er is zo te zien inderdaad wat overlap, maar volgens mij zijn ze beide wel nuttig om te mergen. De cleanup die hierin zit zit volgens mij (nog) niet in de bowtie PR.

jjkester:
I'll wait for information on @i3anaan's PR until tomorrow 12:00.

i3anaan:
De analyze zit zoals eerder gezegd ook in de bowtie PR, maar de rest niet. Ik zou beiden mergen.

PR #91, opened 2015-05-11
FunctionMenu and bugFix

TheMSB:
Created the base for the FunctionMenu, as such MainMenu has been removed as FunctionMenu will replace it. Additionally a window and OSX crashing bug has been removed from Main.java


TheMSB:
If you prefer I can factor out the bug fix consisting of 3 lines.

Code/src/main/java/nl/utwente/group10/ui/Main.java
```diff
@@ -31,8 +31,8 @@ public void start(Stage stage) throws Exception {
         // Init TactilePane
         CustomUIPane tactilePane = new CustomUIPane(catalog);
         tactilePane.setBordersCollide(true);
-        tactilePane.setMinWidth(50000);
-        tactilePane.setMinHeight(50000);
+        tactilePane.setMinWidth(3000);
+        tactilePane.setMinHeight(3000);
```
wandernauta: I'm not sure this is right: the minWidth/minHeight should be equal to the maxWidth/maxHeight if we want to prevent JavaFX from resizing the pane to its contents. You should probably either change both, or neither.

Code/src/main/java/nl/utwente/group10/ui/menu/FunctionMenu.java
```diff
@@ -0,0 +1,142 @@
...
+ * FunctionMenu is constructed out of three different spaces:
+ * {@code searchSpace}, {@code categorySpace} and {@code utilSpace}. The
+ * {@code searchSpace} should be used to display search components that can be
+ * used to find stored functions more quickly, {@code categorySpace} contains an
+ * {@linkplain Accordion} where each category of functions is displayed in their
+ * own {@linkplain TitledPane}. {@code utilSpace} can then contain any form of
+ * utility methods or components that might need quick accessing.
+ * </p>
+ */
+@SuppressWarnings("restriction")
```
wandernauta: What does this `SuppressWarnings` do?

wandernauta:
The code doesn't compile on my machine - seems you're importing a class you've deleted. Check Travis for the file/line. Without that import, though, it seems to work fine.

Code looks very reasonable. I do have a few interface/desgin/interaction questions, though:

- The whole menu is perhaps a bit large at the default zoom level. I don't know how large it will be on the 3M, but I believe the touchscreen has a pretty low dpi (pretty large pixels) so it would be even larger there. At the same time, the drag border to move the menu is quite small. Maybe the font size could be turned down a notch and the border padding increased a bit.
- Scrolling is the fastest way to browse through the menu's, however, if you scroll too much the event gets passed to the pane, which interprets it as a scroll. This could be a bit surprising.
- The whole menu is animated on my machine, slooowly expanding and collapsing. This might be a default thing in JavaFX, and it's fine if you leave it in, but my personal preference would be not to animate it at all (or make the animation much quicker).
- Every right-click creates a new menu, even if you already have one open. This might be on purpose, but it does mean that I can't just right-click somewhere far away to get the menu out of the way. Again, personal preference, just throwing it out there.

Anyway, good work pardner :shipit:

i3anaan:
> Every right-click creates a new menu, even if you already have one open. This might be on purpose, but it does mean that I can't just right-click somewhere far away to get the menu out of the way. Again, personal preference, just throwing it out there.

A persistent menu works a lot better on touch, so I'm happy with this feature!

TheMSB:
> The whole menu is perhaps a bit large at the default zoom level. I don't know how large it will be on the 3M, but I believe the touchscreen has a pretty low dpi (pretty large pixels) so it would be even larger there. At the same time, the drag border to move the menu is quite small. Maybe the font size could be turned down a notch and the border padding increased a bit.

This should be easy enough to change using CSS since all attributes are modifiable.
I am hoping @Kristel-92  will work on this during Sprint 9.

> Scrolling is the fastest way to browse through the menu's, however, if you scroll too much the event gets passed to the pane, which interprets it as a scroll. This could be a bit surprising.

Interesting, I am not sure what causes this. If you have a suggestion for a fix feel free to share.

> The whole menu is animated on my machine, slooowly expanding and collapsing. This might be a default thing in JavaFX, and it's fine if you leave it in, but my personal preference would be not to animate it at all (or make the animation much quicker).

I removed the animation, it should be instantaneous right now.

> Every right-click creates a new menu, even if you already have one open. This might be on purpose, but it does mean that I can't just right-click somewhere far away to get the menu out of the way. Again, personal preference, just throwing it out there.

This is indeed intended behaviour for multi user purposes.

Code/src/main/java/nl/utwente/group10/ui/menu/FunctionMenu.java
TheMSB: It's a check style override, forgot to remove it.

Code/src/main/java/nl/utwente/group10/ui/Main.java
TheMSB: Changed both.

TheMSB:
If anyone finds this interesting: http://stackoverflow.com/questions/21451386/titled-pane-css-settings
Each category 'list' is in fact a TitledPane so in order to modify the visual representation you can use these settings.

wandernauta:
I've taken a better look and I think that you should just be able to catch and consume the scroll event inside `FunctionMenu`, so before it reaches `CustomUIPane`. In other words, you'd throw something like this in the FunctionMenu constructor:

    this.addEventHandler(ScrollEvent.SCROLL, Event::consume);

TheMSB:
@wandernauta, Is there anything left you want changed?

wandernauta:
It looks great, thanks! Merged.

Issue #87, opened 2015-05-10
Types not propagated sideways


wandernauta:
![bug3](https://cloud.githubusercontent.com/assets/126832/7555728/988a3e1c-f754-11e4-9d72-2262e956532c.png)

The second argument for (==) should show Float, not (Eq a).

PR #92, opened 2015-05-13
Basic backend implementation of user defined functions

jjkester:
Feedback please!

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
@@ -7,9 +7,7 @@ import nl.utwente.group10.haskell.hindley.HindleyMilner;
 import nl.utwente.group10.haskell.type.FuncT;
 import nl.utwente.group10.haskell.type.Type;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 
 /**
  * Represents a user-defined Haskell function. Holds arguments which can be used for building the function.
@@ -17,75 +15,60 @@ import java.util.Optional;
 public class Function extends Expr {
 
     /**
-     * An argument for a function. Can be used as a variable in a function definition.
+     * An argument for a function. A FunctionArgument can be used as a variable in a function definition.
      */
-    public class FunctionArgument extends Expr {
+    public static class FunctionArgument extends Expr {
+        /** The type for this argument. */
+        private Type type;
 
-        /** The function this argument is for. */
-        private Function function;
+        /** The name of this argument. Will be randomly generated to a unique value. */
+        private String name;
 
         /**
-         * @param function The function this argument is for.
+         * Constructor for a FunctionArgument with an explicit type.
+         *
+         * @param type The required type for this argument.
          */
-        protected FunctionArgument(Function function) {
-            this.function = function;
+        public FunctionArgument(Type type) {
+            this.type = type;
+            this.name = String.format("var_%s", UUID.randomUUID().toString().substring(0, 8));
         }
 
         /**
-         * @return The 0-indexed position of this argument in the function definition or -1 if the argument is not used.
+         * Constructor for a FunctionArgument without an explicit type. Using this constructor is recommended.
          */
-        public int getPosition() {
-            return this.function.arguments.indexOf(this);
+        public FunctionArgument() {
+            this(HindleyMilner.makeVariable());
         }
 
         @Override
         public Type analyze(Env env, GenSet genSet) throws HaskellException {
-            return HindleyMilner.makeVariable();
+            return this.type;
         }
 
         @Override
         public String toHaskell() {
-            return String.format("%s%d", Function.ARG_PREFIX, this.getPosition());
+            return this.name;
         }
 
         @Override
         public String toString() {
-            return String.format("%s%d_%s%d", Function.FUNC_PREFIX, this.function.hashCode(), Function.ARG_PREFIX, this.getPosition());
+            return this.name;
         }
     }
-
-    /** The prefix for the function part of function argument names. */
-    protected static final String FUNC_PREFIX = "arg";
-
-    /** The prefix for the argument part function argument names. */
-    protected static final String ARG_PREFIX = "arg";
 
     /** The arguments for this function. */
     private List<FunctionArgument> arguments;
 
     /** The expression that forms the base of this function. */
-    private Optional<Expr> expr;
+    private Expr expr;
 
     /**
      * @param expr The expression that forms the base of this function.
      */
-    public Function(Expr expr) {
-        this.arguments = new ArrayList<>();
-        this.expr = Optional.ofNullable(expr);
-    }
-
-    /**
-     */
-    public Function() {
-        this(null);
-    }
-
-    /**
-     * Sets the expression for this function.
-     * @param expr The expression to set.
-     */
-    public void setExpr(Expr expr) {
-        this.expr = Optional.ofNullable(expr);
+    public Function(Expr expr, FunctionArgument ... arguments) {
+        this.arguments = Arrays.asList(arguments);
+        this.expr = expr;
     }
 
     /**
@@ -96,37 +79,12 @@ public class Function extends Expr {
         return this.arguments.toArray(args);
     }
 
-    /**
-     * Adds a new argument to this function.
-     * @return The new argument.
-     */
-    public FunctionArgument addArgument() {
-        FunctionArgument arg = new FunctionArgument(this);
-        this.arguments.add(arg);
-        return arg;
-    }
-
-    /**
-     * Removes the given argument from this function. The argument is not removed from any expressions within this
-     * function.
-     * @param arg The argument to remove.
-     */
-    public void removeArgument(FunctionArgument arg) {
-        this.arguments.remove(arg);
-    }
-
     @Override
     public Type analyze(Env env, GenSet genSet) throws HaskellException {
-        Type type;
+        Type type = this.expr.analyze(env).prune().getFresh();
 
-        if (this.expr.isPresent()) {
-            type = this.expr.get().analyze(env, genSet);
-        } else {
-            type = HindleyMilner.makeVariable();
-
-            for (int i = this.arguments.size(); i > 0; i--) {
-                type = new FuncT(this.arguments.get(i-1).analyze(env, genSet), type);
-            }
+        for (int i = this.arguments.size(); i > 0; i--) {
+            type = new FuncT(this.arguments.get(i - 1).analyze(env).getFresh(), type);
         }
 
         return type;
@@ -136,7 +94,7 @@ public class Function extends Expr {
     public String toHaskell() {
         StringBuilder out = new StringBuilder();
 
-        if (this.expr.isPresent()) {
+        if (!this.arguments.isEmpty()) {
             out.append("\\");
 
             for (FunctionArgument argument : this.arguments) {
@@ -144,7 +102,9 @@ public class Function extends Expr {
             }
 
             out.append(" -> ");
-            out.append(this.expr.get().toHaskell());
+            out.append(this.expr.toHaskell());
+        } else {
+            out.append(this.expr.toHaskell());
         }
 
         return out.toString();
@@ -154,16 +114,17 @@ public class Function extends Expr {
     public String toString() {
         StringBuilder out = new StringBuilder();
 
-        if (this.expr.isPresent()) {
+        if (!this.arguments.isEmpty()) {
             out.append("λ");
 
             for (FunctionArgument argument : this.arguments) {
-                out.append(" ").append(argument.toString());
+                out.append(" ").append(argument.toHaskell());
             }
 
-            out.append(" = ");
-            out.append(this.expr.get().toString());
+            out.append(" -> ");
         }
+
+        out.append(this.expr.toHaskell());
 
         return out.toString();
     }
--- a/Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
@@ -1,0 +1,167 @@
+package nl.utwente.group10.haskell.expr;
+
+import nl.utwente.group10.haskell.env.Env;
+import nl.utwente.group10.haskell.exceptions.HaskellException;
+import nl.utwente.group10.haskell.hindley.GenSet;
+import nl.utwente.group10.haskell.hindley.HindleyMilner;
+import nl.utwente.group10.haskell.type.FuncT;
+import nl.utwente.group10.haskell.type.Type;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Represents a user-defined Haskell function. Holds arguments which can be used for building the function.
+ */
+public class Function extends Expr {
+
+    /**
+     * An argument for a function. Can be used as a variable in a function definition.
+     */
+    public class FunctionArgument extends Expr {
+
+        /** The function this argument is for. */
+        private Function function;
+
+        /**
+         * @param function The function this argument is for.
+         */
+        protected FunctionArgument(Function function) {
+            this.function = function;
+        }
+
+        /**
+         * @return The 0-indexed position of this argument in the function definition.
+         */
+        public int getPosition() {
+            return this.function.arguments.indexOf(this);
+        }
+
+        @Override
+        public Type analyze(Env env, GenSet genSet) throws HaskellException {
+            return HindleyMilner.makeVariable(); // TODO Make sure that this is correctly unified and/or make this editable.
+        }
+
+        @Override
+        public String toHaskell() {
+            return String.format("%s%d", Function.ARG_PREFIX, this.getPosition());
+        }
+
+        @Override
+        public String toString() {
+            return String.format("%s%d", Function.ARG_PREFIX, this.getPosition());
+        }
+    }
+
+    /** The prefix for function argument names. */
+    protected static final String ARG_PREFIX = "arg";
+
+    /** The arguments for this function. */
+    private List<FunctionArgument> arguments;
+
+    /** The expression that forms the base of this function. */
+    private Optional<Expr> expr;
+
+    /**
+     * @param expr The expression that forms the base of this function.
+     */
+    public Function(Expr expr) {
+        this.arguments = new ArrayList<>();
+        this.expr = Optional.ofNullable(expr);
+    }
+
+    /**
+     */
+    public Function() {
+        this(null);
+    }
+
+    /**
+     * Sets the expression for this function.
+     * @param expr The expression to set.
+     */
+    public void setExpr(Expr expr) {
+        this.expr = Optional.ofNullable(expr);
+    }
+
+    /**
+     * @return An ordered array of the arguments for this function.
+     */
+    public FunctionArgument[] getArguments() {
+        FunctionArgument[] args = new FunctionArgument[this.arguments.size()];
+        return this.arguments.toArray(args);
+    }
+
+    /**
+     * Adds a new argument to this function.
+     * @return The new argument.
+     */
+    public FunctionArgument addArgument() {
+        FunctionArgument arg = new FunctionArgument(this);
+        this.arguments.add(arg);
+        return arg;
+    }
+
+    /**
+     * Removes the given argument from this function. The argument is not removed from any expressions within this
+     * function.
+     * @param arg The argument to remove.
+     */
+    public void removeArgument(FunctionArgument arg) {
+        this.arguments.remove(arg);
+    }
+
+    @Override
+    public Type analyze(Env env, GenSet genSet) throws HaskellException {
+        Type type;
+
+        if (this.expr.isPresent()) {
+            type = this.expr.get().analyze(env, genSet);
+        } else {
+            type = HindleyMilner.makeVariable();
+
+            for (int i = this.arguments.size(); i > 0; i--) {
+                type = new FuncT(this.arguments.get(i-1).analyze(env, genSet), type);
+            }
+        }
+
+        return type;
+    }
+
+    @Override
+    public String toHaskell() {
+        StringBuilder out = new StringBuilder();
+
+        if (this.expr.isPresent()) {
+            out.append("\\");
+
+            for (FunctionArgument argument : this.arguments) {
+                out.append(" ").append(argument.toHaskell());
+            }
+
+            out.append(" = ");
+            out.append(this.expr.get().toHaskell());
+        }
+
+        return out.toString();
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder out = new StringBuilder();
+
+        if (this.expr.isPresent()) {
+            out.append("\\");
+
+            for (FunctionArgument argument : this.arguments) {
+                out.append(" ").append(argument.toString());
+            }
+
+            out.append(" = ");
+            out.append(this.expr.get().toString());
+        }
+
+        return out.toString();
+    }
+}
--- a/Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
@@ -32,7 +32,7 @@ public class Function extends Expr {
         }
 
         /**
-         * @return The 0-indexed position of this argument in the function definition.
+         * @return The 0-indexed position of this argument in the function definition or -1 if the argument is not used.
          */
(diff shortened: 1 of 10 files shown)


wandernauta:
Hey there!

This looks pretty good, spot on code-wise as always. I do have some questions/feedback, though, I'll just add those as line notes.

At any rate, great work.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,167 @@
...
+
+        /**
+         * @param function The function this argument is for.
+         */
+        protected FunctionArgument(Function function) {
+            this.function = function;
+        }
+
+        /**
+         * @return The 0-indexed position of this argument in the function definition.
```
wandernauta: ...or `-1` if it's not in there for some reason.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        @Override
+        public Type analyze(Env env, GenSet genSet) throws HaskellException {
+            return HindleyMilner.makeVariable(); // TODO Make sure that this is correctly unified and/or make this editable.
+        }
+
+        @Override
+        public String toHaskell() {
+            return String.format("%s%d", Function.ARG_PREFIX, this.getPosition());
```
wandernauta: This will likely not be unique when lambda's get nested somehow. I don't see how they would, but I imagine they could, and that could lead to some human confusion:

    (\arg0 -> map (\arg0 -> arg0) arg0) 

If you drop the function's hashCode in there somewhere, you'll have a way of differentiating all those arg0's.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,167 @@
...
+        public String toString() {
+            return String.format("%s%d", Function.ARG_PREFIX, this.getPosition());
+        }
+    }
+
+    /** The prefix for function argument names. */
+    protected static final String ARG_PREFIX = "arg";
+
+    /** The arguments for this function. */
+    private List<FunctionArgument> arguments;
```
wandernauta: This seems to mirror the front-end idea that a function application (or here, definition) can have multiple arguments.

I'm not quite sure yet, but my gut tells me it might be easier to go with what FuncT does and make every defined Function have exactly one input and exactly one output, no less and certainly no more, primarily to avoid copying the bug from #87 over to the backend.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,167 @@
...
+    public Function(Expr expr) {
+        this.arguments = new ArrayList<>();
+        this.expr = Optional.ofNullable(expr);
+    }
+
+    /**
+     */
+    public Function() {
+        this(null);
+    }
```
wandernauta: I'm not sure if it would work, but I'd rather see an immutable Function representation that's lightweight enough to build anew every time we do `asExpr`.

It seems that this Function is designed to be mutable and have its state changed when needed, even starting out in a kinda-sorta valid state. All the other subclasses of Expr are pretty much ready to roll and immutable once they're created (with the exception of Value's `setValue`, perhaps from an oversight). So I'd say, make this one immutable as well:

- Drop no-arg constructor: have to know expr *and number of args* before turning them into a Function.
- No setters or list manipulation on arguments.
- Make some things `final` for good measure, like in Apply.
- No real need for `getArguments()` I think - a `getArgument(i)` would perhaps be useful though.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,167 @@
...
+
+        if (this.expr.isPresent()) {
+            type = this.expr.get().analyze(env, genSet);
+        } else {
+            type = HindleyMilner.makeVariable();
+
+            for (int i = this.arguments.size(); i > 0; i--) {
+                type = new FuncT(this.arguments.get(i-1).analyze(env, genSet), type);
+            }
+        }
```
wandernauta: I'm not sure I understand what this does.

If the function definition has a function body, we analyze the body, which probably contains arguments which get inferred along with that analyze, then return that. Makes sense.

But if we don't have a body, we do something else entirely - going through the arguments... backwards? I think?

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,167 @@
...
+        if (this.expr.isPresent()) {
+            out.append("\\");
+
+            for (FunctionArgument argument : this.arguments) {
+                out.append(" ").append(argument.toHaskell());
+            }
+
+            out.append(" = ");
+            out.append(this.expr.get().toHaskell());
+        }
```
wandernauta: This is not entirely correct, lambda syntax in Haskell is `(\x -> x + 10)`, notice no space after backslash (although that may be allowed, I'm not sure) and arrow instead of equals sign.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,167 @@
...
+
+            for (FunctionArgument argument : this.arguments) {
+                out.append(" ").append(argument.toString());
+            }
+
+            out.append(" = ");
+            out.append(this.expr.get().toString());
+        }
+
+        return out.toString();
```
wandernauta: A bit of duplicated code here it seems. Just call `toHaskell()`, it'll be fine.

Code/src/test/java/nl/utwente/group10/haskell/expr/FunctionTest.java
```diff
@@ -0,0 +1,64 @@
...
+    public void testAnalyze() throws HaskellException {
+        Function add5 = new Function();
+        Expr add5Expr = new Apply(new Ident("(+)"), add5.addArgument());
+        add5.setExpr(add5Expr);
+
+        assertEquals("\\ arg0 = ((+) arg0)", add5.toHaskell());
+
+        Expr addition = new Apply(add5, new Value(new ConstT("Int"), "5"));
+
+        assertEquals("Int", addition.analyze(this.env).prune().toHaskellType());
```
wandernauta: I might just be undercaffeinated, but I'm not sure I also get Int in ghci:

````
Prelude> :t (\x -> (+) x)
(\x -> (+) x) :: Num a => a -> a -> a
Prelude> :t (\x -> (+) x) (5 :: Int)
(\x -> (+) x) (5 :: Int) :: Int -> Int
Prelude> :t (\x -> (+) x) (5 :: Int) (10 :: Int)
(\x -> (+) x) (5 :: Int) (10 :: Int) :: Int
Prelude> (\x -> (+) x) 5 10
15
````

**Comment by wandernauta** [comment]
And that's it! Hope that helps.

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 1 above)
Comment (jjkester) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: Good point.

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 2 above)
Comment (jjkester) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: Good idea!

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 3 above)
Comment (jjkester) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: Will think about that.

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 4 above)
Comment (jjkester) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: This was indeed my first plan. However, I soon encountered an interesting issue.

The function consturctor requires you to set the expression. However, the expression is built using the functions arguments, which are only available after creating the function. (What came first, the chicken or the egg?)

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 5 above)
Comment (jjkester) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: If there is no body, the type of the function is something like a -> b -> c depending on the number of arguments.

The loop is backwards to make sure that the last argument is at the bottom of the FuncT tree. Otherwise the result would be c -> b -> a while a -> b -> c was intended.

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 6 above)
Comment (jjkester) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: The space is allowed, as are multiple arguments (examples of both are somewhere on https://www.haskell.org/tutorial/functions.html).

The arrow/equals sign is clearly a bug.

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 7 above)
Comment (jjkester) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: Same reasons as with other Expr subclasses. The methods are different now (but only marginally).

File: Code/src/test/java/nl/utwente/group10/haskell/expr/FunctionTest.java (hunk 8 above)
Comment (jjkester) [inline Code/src/test/java/nl/utwente/group10/haskell/expr/FunctionTest.java]: This is indeed wrong (and may be the most complicated part of this thing). The Apply should have set x to 5, but instead it added it to the end. (I thought what I did was too easy - turns out it is.)

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 4 above)
Comment (wandernauta) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: Hmm, yes, very true. 

Here's an idea. I don't like it very much myself, but maybe it sparks some inspiration. You could allow frontend to build FunctionArgument exprs all by itself, making FunctionArgument basically nothing more than a wrapped integer *n*. A FunctionArgument that wraps a 1 would mean 'the second argument to a function, whatever it's going to be'. You'd collect these things at `Function` construction time, count them (the function will have `max(n)+1` args), attach some types, do some inference and so on and so forth.

Practically this would put the egg before the chicken. Does that make sense?

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java (hunk 5 above)
Comment (wandernauta) [inline Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java]: Ah, right. But do you need N calls to `analyze()`? Wouldn't (shouldn't) a single `analyze()` at the end of the loop do the trick?

Also, instead of walking backwards through the loop, you might get away with swapping the arguments, [like this](https://github.com/wandernauta/Groep10/blob/master/Code/src/main/java/nl/utwente/group10/ui/components/blocks/FunctionBlock.java#L216-L223). (I know that's Apply, but FuncT should be similar I think).

**Comment by jjkester** [comment]
Current state is broken due to a failing unit test.

The fix for this unit test has something to do with upward type propagation and *will* need implementing one or more extra HM rules.

**Comment by wandernauta** [comment]
But you would like to merge as-is?

**Comment by jjkester** [comment]
Of course not. I want to fix the issue because this issue prevents us from knowing the type of a user defined function. Which is kind of a big deal I guess.

**Comment by jjkester** [comment]
FunctionArguments now require an explicit type. This should avoid a lot of work :-). If you're happy to merge, I am as well.

**Comment by wandernauta** [comment]
Thanks! I'll take a look.

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java [hunk 9]
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public static class FunctionArgument extends Expr {
+        /** The type for this argument. */
+        private Type type;
+
+        /** The name of this argument. Will be randomly generated to a unique value. */
+        private String name;
+
+        /**
+         * Constructor for a FunctionArgument with an explicit type.
```
wandernauta: What does 'explicit' mean here? `type` could still be a `new VarT()`, not a lot of explicitness about that.

Could you require `type` to be a tree of constant types perhaps?

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public Function(Expr expr, FunctionArgument ... arguments) {
+        this.arguments = Arrays.asList(arguments);
+        this.expr = expr;
+    }
+
+    /**
+     * @return An ordered array of the arguments for this function.
+     */
+    public FunctionArgument[] getArguments() {
```
wandernauta: Why return an array here? Why not just a `List`?

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    public Type analyze(Env env, GenSet genSet) throws HaskellException {
+        Type type = this.expr.analyze(env).prune().getFresh();
+
+        for (int i = this.arguments.size(); i > 0; i--) {
+            type = new FuncT(this.arguments.get(i - 1).analyze(env).getFresh(), type);
+        }
+
+        return type;
```
wandernauta: I feel like this does too many rounds of `analyze()` and too many rounds of `getFresh()`, but that's just a feeling, I've got nothing to back it up. Will investigate.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            out.append(" -> ");
+            out.append(this.expr.toHaskell());
+        } else {
+            out.append(this.expr.toHaskell());
+        }
+
+        return out.toString();
+    }
```
wandernauta: Other option:

````
    @Override
    public String toHaskell() {
        if (!this.arguments.isEmpty()) {
            Iterable<String> argNames = Iterables.transform(this.arguments, FunctionArgument::toHaskell);
            return String.format("\\ %s -> %s", Joiner.on(" ").join(argNames), this.expr.toHaskell());
        } else {
            return this.expr.toHaskell();
        }
    }
````

Not necessarily better, just a bit shorter.

File: Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java [hunk 13]
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    public String toString() {
+        StringBuilder out = new StringBuilder();
+
+        if (!this.arguments.isEmpty()) {
+            out.append("λ");
```
wandernauta: An actual lambda symbol! I like that.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -62,6 +67,30 @@ public String toHaskellType() {
...
+        return new ConstT(this.constructor, this.getFreshArgs());
+    }
+
+    /**
+     * Returns an array of fresh arguments. Selectively calls {@code getFresh} on each argument. When the same Type
+     * instance appears multiple times in the arguments no new type is instantiated. Instead, the fresh type is reused.
+     * @return An array of fresh arguments.
+     */
+    protected Type[] getFreshArgs() {
+        List<Type> fresh = new LinkedList<>();
```
wandernauta: Any specific reason to go with `LinkedList` here? It's not wrong, I'm just curious.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -62,6 +67,30 @@ public String toHaskellType() {
...
+
+        for (Type arg : this.args) {
+            if (fresh.contains(arg)) {
+                fresh.add(fresh.get(fresh.indexOf(arg)));
```
wandernauta: Wouldn't `fresh.get(fresh.indexOf(arg))` just be `arg`?

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
jjkester: Enforcing that would require to look through the whole tree - I think that this should not be the responsibility of the class.

Also, you might want (for some reason) to explicitly have a VarT - for example if your lambda function works with all ```Num``` or ```Eq```.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
jjkester: Mutability/immutability. To be fair, the arguments could be stored internally as an array as they are no longer mutable. I think I will go with that (unless you have another opinion, then please let me know).

Code/src/main/java/nl/utwente/group10/haskell/type/FuncT.java
```diff
@@ -30,4 +30,10 @@ public final String toHaskellType() {
         out.append(")");
         return out.toString();
     }
+
+    @Override
+    public FuncT getFresh() {
+        Type[] args = getFreshArgs();
+        return new FuncT(args[0], args[1]);
+    }
```
wandernauta: I'm having second thoughts on the FooT classes. I originally thought of them as just convenient places to put some shorthand constructors (`new FuncT(intT, intT)` reads better than `new ConstT("->", intT, intT)`, but the identity of these objects has since been used to decide on the string representation, and I believe frontend does an `instanceof` check somewhere that doesn't really gel with that.

Would it make sense to *not* overwrite getFresh() for every class, throw around plain old ConstT's instead, and make the toString on ConstT itself smart enough to handle the special cases?

(The same goes for ListT and TupleT of course.)

Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
```diff
@@ -7,7 +7,7 @@
 /**
  * Abstract class for Haskell types. Provides an interface for common methods.
  */
-public abstract class Type extends HaskellObject implements Comparable<Type> {
+public abstract class Type extends HaskellObject implements Comparable<Type>, Cloneable {
```
wandernauta: Where's the override for `clone`?

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
jjkester: I think it works as follows:

- The ```expr``` needs to be fresh (due to the mutability of ```analyze```. To make sure that we do not have to copy a massive tree, its type is pruned.
- The arguments need to be fresh as well, otherwise the ```Num a``` of an argument could be transformed to ```Int``` for all usages of the function.

Remember, while we specify it as a lambda it will be used more than once. So this ```analyze()``` is in charge of making sure that *usage* will not change the inner expression.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
jjkester: Me too. I did have it in at first, but somewhere in many undo's it got lost in my initial commit. But it's back!

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
jjkester: Me too. I might have written this in a partially awake state. However, a linked list might be efficient as there usually are one or two arguments. (This is not the case with tuples, but since we are only supporting 2-tuples, it might ultimately be true for tuples as well.)

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
wandernauta: You could go for an ImmutableList to make the intent clear, if you want to.

If I understand correctly, arrays are mutable (you can replace them with another array), the collection they represent is mutable (you can replace elements with other elements) and their elements also mutable (they can change their own state), it's just that arrays won't resize.

On a `final ImmutableList`, you've prevented yourself from the first two things. The last thing is still possible - good thing too, because that's what `analyze` does.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
jjkester: No. I had to look at it for a good five minutes to understand why it works. Conclusion is it *almost* works.

```LinkedList.contains``` and ```LinkedList.indexOf``` use ```Object.equals```. Since a fresh argument is equal to its source (this is a mistake - this is not true for ```VarT```), it would return the fresh equivalent instead of the argument itself. I thought it to be quite clever since a map is not needed.

I'm going to refactor this into a map so it works for ```VarT```s as well.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
wandernauta: Hmm, that's quite possible. However, at the end of the method you call toArray on the whole thing, so... I don't know, I'll leave it up to you.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
wandernauta: Ooh, clever! I think. I always have a hard time wrapping my brain around `equals`...

Code/src/main/java/nl/utwente/group10/haskell/type/FuncT.java
jjkester: It would be very un-Java-like. Furthermore, the responsibility of toHaskell and toString would be in the wrong place - I would be doing ```instanceof``` in ```ConstT``` for its subclasses. That is not nice and does certainly not improve extendibility. So I'm not a big fan of it only for the sake of simplifying its subclasses. If the subclasses are merely a convenience, a factory method would be better anyways.

Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
jjkester: That interface is no longer needed. Its just a sad, lonely survivor of my many tries at this.

Code/src/main/java/nl/utwente/group10/haskell/type/FuncT.java
wandernauta: Hmm, no, I don't think you would do instanceof in `ConstT.toString`, that would be the wrong way around: the idea is to make the object identity less important, not more important. You'd switch on `this.constructor` instead, I think.

The current set-up would be very Java-like if ConstT was abstract, but it isn't, so you have FuncT's which are somehow more special but still compare equal to plain ConstT's:

````
    @Test
    public final void testNaked() {
        ConstT u = new ConstT("Unit");

        // A function type is a constant type with a constructor of ->
        ConstT f = new ConstT("->", u, u);

        // You can also write that as new FuncT(u, u)
        FuncT ft = new FuncT(u, u);

        // As you can see, these two types are completely equal (compareTo == 0)
        assertEquals(0, f.compareTo(ft));

        // Except that they aren't
        assertEquals(f.toHaskellType(), ft.toHaskellType());
    }
````

The factory method sounds like a neat idea. You could have it do some FuncT-specific trickery as well.

Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
wandernauta: Very recognizable :wink:

Code/src/main/java/nl/utwente/group10/haskell/type/FuncT.java
jjkester: I think that IF we are going to do this, it belongs in a new PR. But I think there are better ways to spend time now.

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
jjkester: I think I'll keep it this way.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
jjkester: I'm aware of that, I think I'll keep it as-is.

Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
jjkester: Because of the ```toArray``` a new object is returned - so there is no way to change the ```Function```, only the elements in it (which should be immutable aside from ```Value```).

wandernauta: Hmm, right. So what you're saying is that `getArguments` will get you a copy - that makes sense.

jjkester:
@wandernauta Could you take a loot at this again? I would like to have it merged (if all is well).

wandernauta:
Of course! Thanks. :+1:

PR #93, opened 2015-05-18
Add zoom button toolbar/overlay

wandernauta:
Also disables scroll-to-zoom on touch, as the one-finger scroll gesture
conflicted with the one-finger pan gesture.

It should be easy to add other buttons as well, if required.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/Main.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/Main.java
@@ -36,8 +36,6 @@ public class Main extends Application {
         tactilePane.setMaxWidth(50000);
         tactilePane.setMaxHeight(50000);
 
-        tactilePane.getStylesheets().add("/ui/style.css");
-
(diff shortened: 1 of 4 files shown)


wandernauta:
2015-05-18-153414_563x475_scrot

Code/src/main/java/nl/utwente/group10/ui/Main.java
```diff
@@ -36,8 +36,6 @@ public void start(Stage stage) throws Exception {
         tactilePane.setMaxWidth(50000);
         tactilePane.setMaxHeight(50000);
 
-        tactilePane.getStylesheets().add("/ui/style.css");
```
jjkester: What did this? Could this break something?

jjkester:
Works perfectly. Only wondering what that stylesheet load did, it does not seem to break anything when removed, so I'll go ahead and merge this.

wandernauta:
I didn't remove the stylesheet - I just moved it to the Scene instead of the tactilePane so that the styles would also apply to the ZoomOverlay (which is the tactilePane's parent).

PR #94, opened 2015-05-18
Add slider and RGB blocks

wandernauta:
The slider block allows inputting floating point values from 0.0 to 1.0 in a
convenient fashion (if you want another range, just add a multiplication or
addition block).

The RGB block displays a triple of floating point values as a colored pane. Of
course, `(0, 0, 0)` is black, while `(1.0, 1.0, 1.0)` is white.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/Main.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/Main.java
@@ -14,6 +14,7 @@ import nl.utwente.group10.ghcj.GhciException;
 import nl.utwente.group10.haskell.catalog.HaskellCatalog;
 import nl.utwente.group10.ui.components.CustomAlert;
 import nl.utwente.group10.ui.components.blocks.DisplayBlock;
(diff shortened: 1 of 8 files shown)


wandernauta:
Note that the slider causes, and the RGBBlock expects, an `invalidateConnectionState` to do an update - this was fine back when it was called `invalidate`, but with the new name it doesn't make a lot of sense anymore.

@i3anaan, is there a more granular way to do this?

wandernauta:
2015-05-18-183154_572x588_scrot

:sparkles:

jjkester:
Waiting on input from @i3anaan, or @wandernauta if he discovers this himself.

wandernauta:
I've thought about this a bit more. I don't believe there's a better way to do this without overhauling a lot of stuff. I would love to see such an overhaul, but I wouldn't block this PR on it.

jjkester:
OK. The rest of it seems almost too simple to be working, but it does! Good job :+1:

PR #95, opened 2015-05-18
Add first version of graph block

wandernauta:
GraphBlocks take a single input: a function from Float to Float (could perhaps
be Floating to Num, I haven't checked). It then draws the results of that function
as a nice-looking line graph by mapping it over a (Haskell-generated) list of
intermediate values.

The horizontal range is fixed for now (from -5 to +5), as is the step size; the
vertical axis will automatically scale with the input data.

This PR also contains a (temporary, I hope) means of changing the bowtie index
on functions with the arrow keys, as without that, it's hard to partially apply
functions.

Like PR #94, this PR depends on `invalidateConnectionState` to do something
that doesn't have anything to do with connection state - I'm not sure how to
fix that.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/CustomUIPane.java
@@ -17,6 +17,7 @@ import nl.utwente.group10.ghcj.GhciSession;
 import nl.utwente.group10.haskell.catalog.HaskellCatalog;
 import nl.utwente.group10.haskell.env.Env;
 import nl.utwente.group10.ui.components.blocks.Block;
+import nl.utwente.group10.ui.components.blocks.FunctionBlock;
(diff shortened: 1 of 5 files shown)


wandernauta:
Woo pretty pictures

2015-05-18-174834_770x642_scrot

jjkester:
Nice that it is working so quickly. Looks great!

However, I do like to avoid having a temporary thing in master. At the same time the change is very minor and the functionality could still be useful after a good way to do this has been implemented.

Conclusion: go ahead and merge ths when you feel like it.

wandernauta:
How would you feel if I merged this without the bowtie kludge? The graph functionality wouldn't really be all that useful until bowties were functional, but we would keep the kludge out of master.

jjkester:
Would be fine I guess. Hopefully it will add pressure to fully implement bowties in the UI.

wandernauta:
Merged! Thanks for the review.

PR #96, opened 2015-05-18
Add inspector window

wandernauta:
This PR adds an 'inspector' window that shows the expression tree as well as
the generated Haskell source for the selected block.

Since #82, CustomUIPane is no longer notified when the program changes, so you
may have to click the 'refresh' button on the Inspector if you change your
program and it doesn't seem to update.

The window is shown when the Z key is pressed.


wandernauta:
2015-05-18-205142_1409x975_scrot

wandernauta:
Investigating failing unit tests.

wandernauta:
Unit tests are green.

jjkester:
Nice feature! Helps debugging the code and with demonstration purposes.

wandernauta:
Thanks for the review!

PR #97, opened 2015-05-18
RFC: First version of conditional block

wandernauta:
The conditional block is a visual representation of the if-then-else construct
in Haskell: if the Bool in the first anchor evaluates to True, the result is
the expression of the second anchor, else the result comes from the third
anchor. Of course, the types of the second and third anchors have to unify.

This first version already has working asExpr and type inference on the backend
(see tests), but I'm not sure I implemented InputBlock/OutputBlock correctly -
type checking on the frontend doesn't seem to work like I expect it to. This
could be related to #87.

I haven't thought about how the thing should look; comments are more than
welcome.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/haskell/expr/Cond.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/expr/Cond.java
@@ -1,0 +1,51 @@
+package nl.utwente.group10.haskell.expr;
+
+import nl.utwente.group10.haskell.env.Env;
+import nl.utwente.group10.haskell.exceptions.HaskellException;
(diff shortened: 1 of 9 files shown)


wandernauta:
2015-05-18-221349_768x566_scrot

Woo more pictures

jjkester:
Looks great! However, I do not think the focus should be on this right now.

@ everyone: **Do not spend too much time on this**. Still awesome though.

wandernauta:
Yeah, I wanted to know how much work it would be to implement a first version of something like this. Apparently not a lot.

wandernauta:
Closing; it would be quite a lot of work to get this into master and it has disappeared from required features lists.

PR #105, opened 2015-06-03
Report backend

jjkester:
Some work on the backend part of the report. Probably not final.

Stuff to add:
- [x] Class diagrams
- [x] ~~Code examples (developer guide)~~
- [x] Diagram Expr-tree to Haskell


wandernauta:
This looks great, thanks! A few minor points, as always:

- There seems to be some duplication between the design and implementation chapters. Now, this is probably because we implemented pretty much exactly what we designed, but perhaps you could replace some mentions with references to make it seem less repetitive. 
- I think the description of our intermediate representation could perhaps use a nice illustration that shows the transition from program to Expr tree to Haskell code.
- The developer guide sounds like a useful addition, but I had imagined the user guide to be more of a chapter to xerox and hand out to new users, i.e. a bit like the tutorial portion of the poster. Would it make sense to have the developer guide as a separate chapter? Or try to put the remarks in the Implementation chapter?

Again, thanks a lot!

jjkester:
- I think that repeating a bit in the implementation chapter is not that bad. The design choices are described in the design chapter, as they should be. The implementation is based on these choices, so I think that repeating a bit of the design choices in the implementation chapter is okay. I do not think it is safe to assume that someone who reads the implementation chapter has read the design chapter. I will look into making more references if it does not hurt the readability of the thing.
- Yes. Added this to the todo-list.
- I mainly created the thing for a more in-depth description for the people that may be continuing working on Viskell. Someone might be interested in the design of the software, but not so much in how to add new types. I leave it up to you to decide where exactly this should be placed (maybe in a separate document?).

jjkester:
The commit is coming ;-)

I decided to drop the developer documentation - maybe the package/class javadoc is a better place for that. (We might have to supply HTML javadoc anyways.)

wandernauta:
I vaguely remember Javadoc was a requirement, and even if not, I think it would be a good idea to do so.

PR #109, opened 2015-06-12
Don't call toString in toHaskell and vice versa

wandernauta:


jjkester:
Quite stupid of me to do it the wrong way around...

Issue #112, opened 2015-06-18
Fixes for UI type labels, faster type inference.


jjkester:
**Proposed enhancements for back end:**
- ```Expr``` should provide both ```analyze``` and ```getType``` (where the latter returns a cached type or calls analyze if there is none). ```analyze``` should call ```getType``` instead of ```analyze``` on related ```Expr``` objects.

**Proposed fixes for front end:**
- Function blocks *need* to base the type for labels on the result of ```analyze```. This result should be calculated recursively through the tree, in an efficient manner (for example by caching the expression of blocks and selectively invalidating these).

--------

This should result in the following behaviour:
- Each UI block calculates its own type.
- Each subsequent UI block calculates its own type with the cached result of the previous calculation.

jjkester:
A better implementation is on the way - closing this for now.

PR #113, opened 2015-06-18
Let Expr cache expressions

jjkester:
Part of #112.

Also fixes InspectorWindow to not (possibly) modify types when UI calls type checker correctly.


jjkester:
This is without the UI stuff.

wandernauta:
Would you like to have this merged into `master`, `demo` or neither?

jjkester:
I think master is fine, as this is just a preparation for UI fixes.

The slight improvement of the inspector window is not that important for the demo either.

wandernauta:
Seems to work, will merge into `master`. Thanks! :+1:

PR #117, opened 2015-06-22
Set minimum window size to 800x600

jjkester:


wandernauta:
I could be wrong, but I do not believe that this is correct. The minimum and maximum size for the tactile pane are set to be exactly equal to prevent the pane from resizing itself depending on its contents. To change the window size, try the minimum and maximum sizes of the Stage (window).

jjkester:
Fixed.

wandernauta:
Looks and works great. Can you give this a quick rebase?

jjkester:
Rebased.

wandernauta:
Marvelous! Thanks.

PR #120, opened 2015-06-26
Massive UI cleanup, refactoring and documentation

i3anaan:
Lots of changes, probably more then I can remember.
- Overall lots of cleanup of code and refactoring and better javaFX usage.
- Documentation for most UI classes
- ConnectionAnchors now have a signature, and will typecheck this themselves.
- Error state checking is now also moved to a property inside ConnectionAnchors.
- Block error state is removed.
- Moved the ConnectionState methods to a ConnectionStateDependent interface.
- ConnectionAnchors now implement the ConnectionStateDependent interface.
- Simplified the InputBlock and OutputBlock interfaces.
- Renamed bowtie and bowtieIndex to knot and knotIndex
- Some small fixes to the UI related to updating.

Things not included in this PR:
- Better Type propagation
- Fix for the Type display issue
- Fix for the CircleMenu
- Cleanup and Documentation for the UI menu classes and the UI (general) classes
- import cleanup


wandernauta:
Whoa, that is a lot of changes. Thanks! I'll have a first stab at reviewing first thing tomorrow.

From a cursory glance, though: Have you talked with @jjkester yet? I don't think the backend utils typeMatch method does what you want, and (without looking at the code) I don't immediately see how moving logic to the anchors would make stuff easier.

i3anaan:
Moving the types to the anchors was definitely easier, and unrelated to the talk I still need to have with @jjkester.
typeMatch() could be replaced after updating how the type propagation in general works.

i3anaan:
To clarify as to why I moved types to the anchors:
Previously the FunctionBlock and InputAnchors played a game of pong with how their methods called eachother, it was a chaotic mess. That is now a lot easier and more logical, the effect of this can be seen in the much simpler FunctionBlock class. This also has almost no effect on the other Blocks, but they can now very easily react to for example type errors if they want.

TheMSB:
@i3anaan : This looks like a huge amount of changes, wouldn't it be better to split up this PR? This seems to initially start out with bow tie implementation and then moves on to do a bunch of other stuff including refactoring in general. I think it will be much easier to evaluate each feature separately, that way if one of the features has issues the rest can still be merged.

Code/src/main/java/nl/utwente/group10/haskell/hindley/HindleyMilner.java
```diff
@@ -39,6 +40,8 @@ public static void unify(final Expr context, final Type t1, final Type t2) throw
         final Type a = t1.prune();
         final Type b = t2.prune();
 
+        HindleyMilner.logger.setLevel(Level.OFF);
```
wandernauta: It doesn't really make sense to do do this every time unify is called. Maybe you could use a static initializer block here?

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
```diff
@@ -31,4 +33,18 @@ public static Type dive(ConstT function, int steps) {
         }
         return function;
     }
+    
+    /**
+     * @return Whether or not t1 and t2 unify with each other.
```
wandernauta: That's not really what this does. Instead, it *changes* t1 to match t2 (or vice versa). It will also recursively change other (seemingly unrelated) things if that's what's needed to make that happen. If there's no way to make it happen, it might *still* change t1 and t2, but it will also throw a type error/return false.

Please, *please* talk with @jjkester or at least look at the diagram I sent you. He will probably also tell you you probably shouldn't call unify from the frontend directly.

Code/src/main/java/nl/utwente/group10/ui/Main.java
```diff
@@ -39,6 +40,9 @@ public void start(Stage stage) throws Exception {
         ValueBlock valueBlock = new ValueBlock(tactilePane);
         DisplayBlock displayBlock = new DisplayBlock(tactilePane);
         tactilePane.getChildren().addAll(valueBlock, displayBlock);
+        
+        
+
```
wandernauta: Whitespace.

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Interface for Nodes who are dependent on their ConnectionState.
+ */
+public interface ConnectionStateDependent {    
+    /**
+     * Tells the Node that its current state (considering connections) possibly has
+     * changed. Default implementation does not react to a potential state
+     * change.
+     *
+     * This method should only be called after the Node's constructor is done.
```
wandernauta: > This method should only be called after the Node's constructor is done.

Doesn't Java guarantee this? You can't call methods on objects you haven't constructed yet.

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
```diff
@@ -0,0 +1,49 @@
...
+     * 
+     * This method will invalidate the Node even if the state did not change.
```
wandernauta: What does 'invalidate' mean here?

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public void invalidateConnectionState();
+    
+    /**
+     * @return The ConnectionState the Node is in.
+     */
+    public int getConnectionState();
+    
+    /**
+     * Does the same as invalidateConnectionState(), but cascading down to other
```
wandernauta: > down

Which direction is down? And are you sure those are the only nodes you have to invalidate?

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
```diff
@@ -0,0 +1,49 @@
...
+     * Nodes which are possibly also (indirectly) affected by the state change.
+     * 
+     * Cascading only happens if this Block is not up-to-date, implying that if
+     * this Block is up-to-date, then so are all following Blocks.
```
wandernauta: > following

Again, which blocks are those? The ones 'below' the current block in screenshots, or the ones above?

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
i3anaan: I have asked previously how to check if types match, and to my knowledge have not received a consise answer, so I did it this way. But again, this PR is not intended to change the type system and the way it is used, as @TheMSB mentioned, this PR is big. Changing all the type stuff would make it bigger.

For the future: I dont care how it happens, I just need a method that can tell me if 2 types match.

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
```diff
@@ -0,0 +1,49 @@
...
+     * This method should only be called after the Node's constructor is done.
+     * 
+     * This method will invalidate the Node even if the state did not change.
+     */
+    public void invalidateConnectionState();
+    
+    /**
+     * @return The ConnectionState the Node is in.
+     */
+    public int getConnectionState();
```
wandernauta: Do you really need this one if you have connectionStateIsUpToDate?

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param state
+     *            The newest visual state
+     */
+    public void invalidateConnectionStateCascading(int state);
+    
+    /**
+     * @return Whether or not the state of this Node confirms to the given newest state.
+     */
+    default public boolean connectionStateIsUpToDate(int state) {
+        return getConnectionState() == state;
```
wandernauta: Wouldn't `>=` make more sense?

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
i3anaan: No it does not, you can call methods from the constructor.

Code/src/main/java/nl/utwente/group10/ui/components/anchors/ConnectionAnchor.java
```diff
@@ -4,48 +4,140 @@
...
+    /** The invisible part of the ConnectionAnchor (the touchZone). */
+    @FXML private Shape invisibleAnchor;
+    
+    /** The signature that is accepted by this anchor. */
+    private Type signature;
+    
+    /** Property storing the error state. */
+    private BooleanProperty isError;
+    
+    /** Property storing the active state. */
```
wandernauta: What does 'active' mean here? Is it the same as TactilePane's active state, that is, 'this node is currently being dragged'?

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
i3anaan: No, if that happened something went horribly wrong.
If something went horribly wrong, I'd rather re-update the Node's state to the current state.

Code/src/main/java/nl/utwente/group10/ui/components/anchors/ConnectionAnchor.java
```diff
@@ -114,17 +206,24 @@ public boolean hasConnection() {
     public boolean isPrimaryConnected() {
         return isConnected(0);
     }
+    
+    /**
+     * @return True if the primary connection is connected, and its types match.
+     */
+    public boolean isPrimaryConnectedCorrect() {
+        return isPrimaryConnected() && BackendUtils.typesMatch(getSignature().getFresh(), getType().getFresh());
```
wandernauta: I don't think this is correct. Please refer to @jjkester.

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
i3anaan: I wanted to make connectionStateIsUpToDate() a default method.
Java interfaces do not support instance fields, so I solved that by adding a getConnectionState().
Now that I think of it, this method can probably be protected (Does an interface support this?).

Removing this would mean that every Class that implements this interface would have to implement isUpToDate

Code/src/main/java/nl/utwente/group10/ui/components/anchors/OutputAnchor.java
```diff
@@ -17,14 +18,25 @@
...
-     *            The parent pane on which this anchor resides.
+     * @param signature
+     *            The Type signature as is accepted by this InputAnchor.
      */
-    public OutputAnchor(Block block, CustomUIPane pane) {
-        super(block, pane);
-        new AnchorHandler(pane.getConnectionCreationManager(), this);
+    public OutputAnchor(Block block, Type signature) {
+        super(block, signature);
+        getInvisibleAnchor().setTranslateY(getInvisibleAnchor().getTranslateY() * -1);
```
wandernauta: What does this line do?

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
jjkester: ```unify()``` should indeed never be called from the front end.

The front end should not test whether types match. The front end should just apply things and call analyze appropriately. Then the type checker will tell you if there is something wrong or not.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
```diff
@@ -0,0 +1,352 @@
...
+ * Custom Pane that displays function's input and output and a knot to be able
+ * to control the amount of inputs that are to be applied (amount of currying).
+ * 
+ * The input arguments are ordered using the layoutX, and the knotIndex is based
+ * on this and the knot's layoutX. Usage of translateX is strictly for visual
+ * improvements.
+ * 
+ * Supports multi-touch interaction.
+ */
+public class ArgumentSpace extends Pane implements ComponentLoader{
```
wandernauta: Missing space before `{`.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
```diff
@@ -0,0 +1,352 @@
...
+     * The index of the knot, all inputs with index higher or equal to the
+     * knot are inactive, creating higher order functions.
+     */
+    private IntegerProperty knotIndex;
+    
+    /** List of input arguments. */
+    private List<InputArgument> leftArguments;
+    
+    /** The argument containing the function's output. */
+    @FXML private Label rightArgument;
```
wandernauta: It's not really an argument, more something of a return type, right?

Code/src/main/java/nl/utwente/group10/ui/components/anchors/ConnectionAnchor.java
jjkester: I do not immediately see what this does, but I do know that this is indeed not the right way around. The front end should never have to use the ```getFresh``` method, and shoudn't. Using this method throws away a lot of useful information about types. Also, as said above, the front end should not have to check whether types do match.

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
i3anaan: Its kind of impossible to display type mismatches if the front-end is not allowed to check for type mismatches.

If the typesMatch() method uses analyze thats fine with me.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
```diff
@@ -0,0 +1,352 @@
...
+     * Constructs an ArgumentSpace belonging to the function block as given.
+     * @param block FunctionBlock to which this ArgumentSpace belongs.
+     */
+    public ArgumentSpace(FunctionBlock block, List<Type> inputSignatures) {
+        this.loadFXML("ArgumentSpace");
+        
+        this.block = block;
+        this.inputID = INPUT_ID_NONE;
+        leftArguments = new ArrayList<InputArgument>();
+        knotIndex = new SimpleIntegerProperty(0);
```
wandernauta: Is 0 the most sensible default?

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
```diff
@@ -0,0 +1,352 @@
...
+     */
+    private IntegerProperty knotIndex;
+    
+    /** List of input arguments. */
+    private List<InputArgument> leftArguments;
+    
+    /** The argument containing the function's output. */
+    @FXML private Label rightArgument;
+    
+    /** InputID of the current action (INPUT_ID_NONE if idle) */
```
wandernauta: What is an InputID?

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
```diff
@@ -0,0 +1,352 @@
...
+    public ObservableValue<? extends Number> getTotalWidthProperty() {       
+        return rightArgument.layoutXProperty().add(rightArgument.translateXProperty()).add(rightArgument.widthProperty()).add(H_GAP);
+    }
+    
+    /**
+     * Binds a region's layoutY property to always be vertically centered in the ArgumentSpace. 
+     * @param region
+     */
+    public void centerLayoutVertical(Region region) {
+        region.layoutYProperty().bind(this.heightProperty().divide(2).subtract(region.heightProperty().divide(2)));
```
wandernauta: This could perhaps use a clarifying comment.

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
jjkester: The right way to do things is to just apply the argument, then run analyze in a try catch block, and display error information when an exception occurs. The exceptions contain a lot of useful information by the way, including the exact expression object where the error occured.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
```diff
@@ -0,0 +1,352 @@
...
+    public List<InputArgument> getInputArguments() {
+        return leftArguments;
+    }
+    
+    /**
+     * @param index
+     * @return The input argument with the index as specified.
+     */
+    public InputArgument getInputArgument(int index) {
+        return leftArguments.get(index);
```
wandernauta: This method could probably be axed (see #114).

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
```diff
@@ -0,0 +1,259 @@
...
+        List<Type> inputSignatures = new ArrayList<>();
+        Type functionSignature = getFunctionSignature();
+        for (int i = 0; i < args.size(); i++) {
+            if (functionSignature instanceof FuncT) {
+                Type inputSignature = BackendUtils.dive((FuncT) functionSignature, i + 1);
+                inputSignatures.add(inputSignature);
+            } else {
+                throw new FunctionDefinitionException();
+            }
+        }
```
wandernauta: Isn't this loop basically doing the same thing as the loop above it?

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
```diff
@@ -0,0 +1,259 @@
...
+        
+        argumentSpace = new ArgumentSpace(this, inputSignatures);
+        argumentSpace.setKnotIndex(getAllInputs().size());
+        
+        ((Pane) this.lookup("#nestSpace")).getChildren().add(argumentSpace);
```
wandernauta: Don't really like `lookup`, would prefer JavaFX binding.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
```diff
@@ -0,0 +1,259 @@
...
+        argumentSpace.knotIndexProperty().addListener(e -> invalidateKnotIndex());
+        
+        // Create an anchor for the result
+        output = new OutputAnchor(this);
+        output.layoutXProperty().bind(outputSpace.widthProperty().divide(2));
+        outputSpace.getChildren().add(output);
+        
+        // Make sure the prefWidth is correctly updated.
+        Pane functionInfo = (Pane) this.lookup("#functionInfo");
```
wandernauta: Same here, would prefer JavaFX binding if possible.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+    
+    /** Returns the name property of this FunctionBlock. */
+    public final String getName() {
+        return name.get();
+    }
+
+    /** Sets the name property of this FunctionBlock. */
+    public void setName(String name) {
+        this.name.set(name);
```
wandernauta: Do we actually need/use this method?

Code/src/main/java/nl/utwente/group10/ui/menu/FunctionMenu.java
```diff
@@ -114,8 +119,6 @@ protected void updateItem(FunctionEntry item, boolean empty) {
         graphBlockButton.setOnAction(event -> addBlock(new GraphBlock(parent)));
 
         // TODO remove once debugging is done
-        Button debugButton = new Button("Error All");
-        debugButton.setOnAction(event -> parent.errorAll());
```
TheMSB: Are we really done with all the debugging? This code should really be the last line to go when we are done, doesn't make sense to throw away debugging tools before we finalise the code.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
```diff
@@ -0,0 +1,259 @@
...
+        return type;
+    }
+
+    @Override
+    public Type getOutputType() {
+        try {
+            return asExpr().analyze(getPane().getEnvInstance()).prune();
+        } catch (HaskellException e) {
+            return getOutputSignature();
+        }
```
wandernauta: This is closer, but still not exactly right, I'm afraid.

Catching HaskellException is a bit scary - maybe catch a more specialized exception instead?

jjkester: That would be recommended since the specialized exceptions contain a lot of useful information.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
```diff
@@ -0,0 +1,352 @@
...
+     */
+    public void invalidateArgumentContent() {
+        invalidateOutputContent();
+        invalidateInputContent();
+    }
+    
+    /**
+     * Method to indicate that the content in the output argument Label is possibly outdated.
+     */
+    public void invalidateOutputContent() {
```
wandernauta: Would prefer this (and invalidateInputContent) to be `private` if possible.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
```diff
@@ -0,0 +1,259 @@
...
+    /** Returns the knot index of this FunctionBlock. */
+    public final Integer getKnotIndex() {
+        return argumentSpace.knotIndexProperty().get();
+    }
+
+    /**
+     * @param index The new knot index for this FunctionBlock.
+     */
+    public final void setBowtieIndex(int index) {
+        if (index >= -1 && index <= getAllInputs().size()) {
```
wandernauta: Would it make sense to move this validation to the ArgumentSpace class?

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/InputArgument.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param block Block to which this InputArgument belongs.
+     * @param signature Type signature as should be accepted by the InputAnchor.
+     */
+    public InputArgument(Block block, Type signature) {
+        inputText = new SimpleStringProperty(signature.toHaskellType());
+        this.loadFXML("InputArgument");
+        
+        inputAnchor = new InputAnchor(block, signature);
+        inputAnchor.layoutXProperty().bind(inputLabel.widthProperty().divide(2));
+        inputAnchor.isErrorProperty().addListener((ObservableValue<? extends Boolean> value, Boolean oldValue, Boolean newValue) -> setError(newValue));
```
wandernauta: The lambda signature is a bit long... Would it make sense to use a method reference here instead?

````
inputAnchor.isErrorProperty().addListener(this::setError); 

// ...

private void setError(ObservableValue<? extends Boolean> value, Boolean oldValue, Boolean newValue) {
    setError(newValue)
}

// ...
````

File: Code/src/main/java/nl/utwente/group10/ui/components/blocks/input/DisplayBlock.java [hunk 28]
```diff
@@ -54,26 +55,28 @@ protected DisplayBlock(CustomUIPane pane, String fxml) {
...
         this.loadFXML(fxml);
 
-        inputAnchor = new InputAnchor(this, pane);
-        anchorSpace.getChildren().add(inputAnchor);
+        inputAnchor = new InputAnchor(this, HindleyMilner.makeVariable());
+        inputAnchor.layoutXProperty().bind(inputSpace.widthProperty().divide(2));
+        inputSpace.getChildren().add(inputAnchor);        
+        
+        //Make sure inputSpace is drawn on top.
+        BorderPane borderPane = (BorderPane) inputSpace.getParent();
```
wandernauta: Would prefer the borderPane to be injected through JavaFX.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/input/InputBlock.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    Type getInputType(int index);
+
+    /**
+     * @return All InputAnchors of the block.
+     */
+    List<InputAnchor> getAllInputs();
+
+    /**
+     * @return Only the active (as specified with by the knot index) inputs.
```
wandernauta: So `getActiveInputs` only makes sense on FunctionBlocks, right? Again, see #114.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/output/ValueBlock.java
```diff
@@ -41,15 +41,21 @@
...
     protected ValueBlock(CustomUIPane pane, String fxml) {
         super(pane);
-
         value = new SimpleStringProperty("5.0");
-        output = new OutputAnchor(this, pane);
-
+        
         this.loadFXML(fxml);
+        try {
+            output = new OutputAnchor(this, new Value(new ConstT("Float"), getValue()).analyze(getPane().getEnvInstance()));
```
wandernauta: Why would you `analyze` a single Value with a known type?

Code/src/main/java/nl/utwente/group10/ui/components/blocks/output/ValueBlock.java
```diff
@@ -41,15 +41,21 @@
...
+        } catch (HaskellException e) {
+            throw new TypeUnavailableException();
+        }
 
         outputSpace.getChildren().add(this.getOutputAnchor());
+        outputSpace.toFront();
+        output.layoutXProperty().bind(outputSpace.widthProperty().divide(2));
```
wandernauta: Why not just center the outputSpace in the FXML?

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
```diff
@@ -27,49 +35,163 @@
...
+    private BooleanProperty isError;
 
+    /** 
+     * Construct a new Connection.
+     * @param pane The Pane this Connection is on.
+     */
     public Connection(CustomUIPane pane) {
         this.pane = pane;
+        this.isError = new SimpleBooleanProperty(false);
+        this.isErrorProperty().addListener((ObservableValue<? extends Boolean> value, Boolean oldValue, Boolean newValue) -> setError(newValue));
```
wandernauta: Again, maybe change this to a method reference.

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
```diff
@@ -27,49 +35,163 @@
...
+        isError.set(error);
+    }
+    
+    /** @return The property describing the error state. */
+    public BooleanProperty isErrorProperty() {
+        return isError;
+    }
 
+    
+    /* GET METHODS */
```
wandernauta: Do you mean getters?

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
```diff
@@ -124,81 +269,73 @@ public boolean tryAddAnchor(ConnectionAnchor anchor, boolean overrideExisting, b
...
+    /**
+     * Removes this Connection, disconnecting its anchors and removing this Connection from the pane it is on.
+     */
+    public final void remove() {
+        disconnect();
+        getPane().getChildren().remove(this);
+    }
+
+    /**
+     * @return True if both anchors are present and their types match.
```
wandernauta: Are you sure? I haven't tested it, but looking at the implementation one would think it would also return true if the anchors are not present/connected.

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
```diff
@@ -124,81 +269,73 @@ public boolean tryAddAnchor(ConnectionAnchor anchor, boolean overrideExisting, b
...
+     */
     public final boolean typesMatch() {
-        // TODO Let this return mismatch information?
-        return !isConnected() || typesMatch(startAnchor.get().getType(), endAnchor.get().getSignature());
+        return !isConnected() || BackendUtils.typesMatch(startAnchor.get().getType(), endAnchor.get().getSignature());
```
wandernauta: I believe this modifies startAnchor's type, which I'd say is a bit scary. At least scary enough to note in the Javadoc.

Code/src/main/java/nl/utwente/group10/ui/components/lines/ConnectionLine.java
```diff
@@ -47,15 +50,21 @@ public void setStartPosition(double xCoord, double yCoord) {
...
         setStartPosition(point.getX(), point.getY());
     }
 
     /**
-     * Sets the end coordinate for this Line object.
+     * Sets the end coordinates for this ConnectionLine object.
      *
-     * @param xCoord
-     * @param yCoord
+     * @param xCoord X coordinate local to parent.
```
wandernauta: What would the parent usually be? The pane?

Code/src/main/java/nl/utwente/group10/ui/handlers/AnchorHandler.java
```diff
@@ -72,10 +94,29 @@ private void inputPressed(int inputId){
...
+     * @param x
+     *            The sceneX coordinate of the input.
+     * @param y
+     *            The sceneY coordinate of the input.
+     */
 	private void inputMoved(int inputId, double x, double y){
 		manager.updateLine(inputId, x, y);
 	}
 	
+	    /**
```
wandernauta: Weird tab indent.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
+.error .block {
+    -fx-border-color: red;
 }
 
-.menu.block #categorySpace TitledPane {
-    -fx-padding: 3px;
+.block .header{
+    -fx-font-size: 20px;
+    -fx-font-weight: bold;
+    -fx-font-family: fantasy;
```
wandernauta: What? `fantasy`? Do we bundle that font?

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
+    -fx-text-fill: black;
 }
 
-.menu.block #utilSpace {
-    -fx-padding: 3px;
-    -fx-spacing: 6px;
+.block .title {
+    -fx-font-family: Titillium;
```
wandernauta: If you move the `-fx-font-family` declaration to `.block`, you might not have to mention it as often.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
+    -fx-text-fill: black;
+    -fx-font-size: 20px;
+    -fx-font-weight: bolder;
```
wandernauta: I'm not sure what `bolder` is, but I believe we only ship bold and regular.

Code/src/main/resources/ui/style.css
```diff
@@ -129,3 +208,9 @@ LineChart {
     -fx-font-weight: bold;
     -fx-background-color: rgba(0, 0, 0, 0.3);
 }
+
+.invisTouchZone {
+}
+
+.leftArgument {
+}
```
wandernauta: These last two don't seem to add much :)

By the way, it seems the whole file has a kind of mixture of tabs and spaces going on - could you fix that?

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
 .connection {
     -fx-stroke-width: 3px;
-    -fx-stroke: white;
+    -fx-stroke: #FCEBB6;
     -fx-fill: null;
 }
 
 .connection.error {
-    -fx-stroke: orange;
+    -fx-stroke: red;
```
wandernauta: Why not orange? I like orange.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
+    -fx-font-size: 25px;
+    -fx-font-weight: normal;
+    -fx-text-fill: black;
 }
-.selected .block {
-    -fx-background-color: skyblue;
+
+.selected .function.block {
+    -fx-background-color: #BA6602;
+	-fx-effect: dropshadow(three-pass-box, rgba(0,0,0,0.8), 10, 0, 0, 0);
```
wandernauta: Should be able to move this to `.selected .block`.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
-    -fx-font-size: 14px;
-    -fx-font-weight: none;
-    -fx-background-color:white;
-    -fx-border-color: black;
-    -fx-border-width: 1;
+.menu.block .title:hover {
+    -fx-font-size: 15px;
+    -fx-font-weight: bold;
+    -fx-font-color: #FCEBB6;
+    -fx-background-color: #926C8B;
```
wandernauta: The first three declarations are probably not needed.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
 
-.menu.block {
-    -fx-font-size: 12px;
+.error .block {
+    -fx-border-color: red;
 }
 
-.menu.block #categorySpace TitledPane {
-    -fx-padding: 3px;
+.block .header{
```
wandernauta: Space before `{`.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
-.function.block .description {
+.menu.block {
     -fx-font-size: 12px;
-    -fx-padding: 0;
+    -fx-background-color: #490A3D;
 }
 
-.function.block .nestSpace {
-    -fx-padding: 15;
+.menu.block .title {
```
wandernauta: Space before `{`.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
+    -fx-font-color: #FCEBB6;
+    -fx-background-color: #926C8B;
 }
 
-.function.block #argumentSpace Label.error {
-    -fx-background-color:pink;
-    -fx-border-color: red;
+
+
+.menu.block .title .arrow-button .arrow{
```
wandernauta: Space before `{`.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
 }
 
-.block .title {
-    -fx-text-fill: #ecf0f1;
+.menu.block .list-cell:odd {
+	-fx-background-color: #C8B6C5;	
+}
+
+.argumentSpace {
+}
```
wandernauta: Might not need this block.

Code/src/main/resources/ui/style.css
```diff
@@ -7,95 +7,173 @@ CustomUIPane {
...
+
+.argumentSpace Label.rightArgument {
+    -fx-padding: 0 5 0 5;
+    -fx-font-family: Titillium;
+    -fx-font-size: 14px;
+    -fx-font-weight: normal;
+    -fx-text-fill: white;
+    -fx-background-color: black;
+    -fx-border-width: 2;
+}
```
wandernauta: Could probably simplify/partially merge these two blocks.

Also, changing the border width makes one slightly larger than the other, right? Is that on purpose?

Code/src/test/java/nl/utwente/group10/ui/components/blocks/DisplayBlockTest.java
```diff
@@ -2,6 +2,7 @@
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import nl.utwente.group10.ui.components.blocks.input.DisplayBlock;
```
wandernauta: If you move the test class as well, you don't need this import.

Code/src/test/java/nl/utwente/group10/ui/components/blocks/FunctionBlockTest.java
```diff
@@ -20,9 +20,10 @@
      */
     @Before
     public void setUp() throws Exception {
+        //TODO: not entirely correct.
```
wandernauta: How so?

Code/src/main/java/nl/utwente/group10/haskell/hindley/HindleyMilner.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
i3anaan: Will not change this for this PR

Code/src/main/java/nl/utwente/group10/ui/Main.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
wandernauta: You didn't receive a concise answer because there is no concise answer.

Here's the thing. Haskell uses global type inference to infer the types of its expressions. The Haskell compiler takes the functions you wrote and the functions in its library, and for every function call, every constant, every *anything* tries to come up with a sensible type for it, one that fits with all the rest of the program. 

To put it another way, the only way to check a program to do type inference on it. This will also, as a side effect, tell you whether or not a program is well-typed: If type inference works, you have a well-typed program. As a bonus, you also have a reasonable idea of what the types are going to be. If type inference fails, you don't have a well-typed program.

This sounds like an NP-complete problem, but joy of joys, there are algorithms available to do this kind of inference. Type inference is quite a complicated subject, but we were able to make it work because we picked one that's well-documented (Hindley-Damas-Milner) and implemented a *subset* of that algorithm for the *subset* of Haskell we target, which is relatively simple. 

The end result is that you have a machine available to you that does just that. It takes a whole program and does type inference on it, leaving you with a typed, checked program. It will even let you know if there's a problem.

JJ and I have been working quite long and hard on thinking it through, getting it to work, and testing it under different circumstances. It pains us that you apparently don't want to use or understand it.

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
i3anaan: Decided to keep it this way

Code/src/main/java/nl/utwente/group10/ui/components/anchors/ConnectionAnchor.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/anchors/ConnectionAnchor.java
i3anaan: Will not be fixed in this PR

Code/src/main/java/nl/utwente/group10/ui/components/anchors/OutputAnchor.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/ConnectionStateDependent.java
wandernauta: You're right, and it seems FunctionBlock's constructor actually does this. Does that mean FunctionBlock's constructor is wrong?

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
i3anaan: Yes, for a Block that has not inputs, this is the only valid index. The FunctionBlock sets this to our default (size - 1) immediately after constructing the ArgumentSpace.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
i3anaan: Hard to explain, did my best. Its an ID for an input. (required to support multi-touch)

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
i3anaan: It does what the method does, I see no need to put an extra comment here.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
i3anaan: Not really, this method returns an InputArgument, not an InputAnchor.

Did remove it though, axing could be Done by merging

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
i3anaan: Apperantly not, I'l postpone this to the type overhaul

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
i3anaan: Doen

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
i3anaan: The FunctionBlock's name label is binded to the nameProperty. To my understanding we then should implement all 3 methods.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
wandernauta: Ah, that makes sense. Thanks.

Code/src/main/java/nl/utwente/group10/ui/menu/FunctionMenu.java
i3anaan: We are done with this part of debugging, as a Block error state is not used anymore. The argument error state can easily be tested by using (+) and (==)

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
i3anaan: Will not change for this PR

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/ArgumentSpace.java
i3anaan: Currently not possible, but these methods have no side effects, so no real problems arise from this.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/InputArgument.java
i3anaan: Did not know this was possible like this, changed.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/input/DisplayBlock.java
i3anaan: I disagree, this BorderPane is only relevant because it is the parent, thus I find it more logical to obtain it by requesting the parent.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/input/InputBlock.java
i3anaan: Depends how you look at it.
Especially since ConnectionAnchors can now be inactive themselves, I think this method is useful to have for all Blocks.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/output/ValueBlock.java
i3anaan: Because at this point I have no idea how to use the backend, posponed for another PR.

Code/src/main/java/nl/utwente/group10/ui/components/blocks/output/ValueBlock.java
i3anaan: Made the outputSpace a BorderPane and centered using that.

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
i3anaan: Done

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
i3anaan: Yes, considered that, went with this

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
i3anaan: You are right, updated

Code/src/main/java/nl/utwente/group10/ui/components/lines/Connection.java
i3anaan: Will not be changed in this PR

Code/src/main/java/nl/utwente/group10/ui/components/lines/ConnectionLine.java
i3anaan: Yes, usually, but nothing in this class is setting its parent to the CustomUIPane, so i didnt mention that.

Code/src/main/java/nl/utwente/group10/ui/handlers/AnchorHandler.java
i3anaan: Done

Code/src/main/resources/ui/style.css
i3anaan: This is a font type

Code/src/main/resources/ui/style.css
i3anaan: Done

Code/src/main/resources/ui/style.css
i3anaan: bolder is more bold than bold, I thought it looked better.

Code/src/main/resources/ui/style.css
i3anaan: Currently I'd say red stands better out against the also orange function blocks. Leaving it red, could discuss this more some other time.

Code/src/main/resources/ui/style.css
i3anaan: Done

Code/src/main/resources/ui/style.css
i3anaan: Done

Code/src/main/resources/ui/style.css
i3anaan: Done

Code/src/main/resources/ui/style.css
i3anaan: Done

Code/src/main/resources/ui/style.css
i3anaan: Done

Code/src/main/resources/ui/style.css
TheMSB: > Why not orange? I like orange.

[I Wanted Orange](https://www.youtube.com/watch?v=hxVPPy5w9NA)

Code/src/main/resources/ui/style.css
i3anaan: Border was unused, removed.
merged css

Code/src/test/java/nl/utwente/group10/ui/components/blocks/DisplayBlockTest.java
i3anaan: Done

Code/src/test/java/nl/utwente/group10/ui/components/blocks/FunctionBlockTest.java
i3anaan: (Int -> Int -> Int) is not the type of the (+) function, although it is not relevant in this testcase

Code/src/main/resources/ui/style.css
i3anaan: Done

Code/src/test/java/nl/utwente/group10/ui/components/blocks/FunctionBlockTest.java
wandernauta: Could you just call your function 'foo' instead of '(+)'? Then you wouldn't have to deal with typeclasses.

i3anaan: That is not in our catalog :(
I wanted to update this along with the other type changes

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
i3anaan: You are describing a different problem than the one the method is supposed to solve.
I dont need to infer the Type anymore, I already have 2 types. I just need to know whether they match or not.

Possibly all of this will change when having talked with @jjkester , but as it stands this is how it is used.

i3anaan:
For all the feedback gotten I either fixed it, or commented why I didnt fix it.
It occured to me that it might be wise to merge Master into this branch myself, do you want me to do this?

Code/src/main/java/nl/utwente/group10/ui/BackendUtils.java
wandernauta: If you already have the inferred type here, that must mean that type inference has succeeded. In other words, if that's the case, this method could just be `return true;`.

wandernauta:
Preferably not: we'd end up with two sets of merge commits (first to merge master into this branch, then to merge this branch back into master). You could rebase this branch on top of master, but that would probably be quite a lot of work for not a lot of gain.

The problem is this. I believe your changes to how the frontend handles type checking make it worse than what we have now. I also think that applying these changes would make it harder, not easier, to do it right. This means I'm reluctant to merge these changes.

The PR also doesn't seem to compile on Travis. When I tested it locally, it at least compiled, but unit tests were failing.

Long story short, if you could isolate the many, many unrelated fixes, refactors and smaller changes into seperate PR's (for example, by cherry-picking the commits) I'd be more than happy to merge them. For the type-related changes, talk to JJ first.

i3anaan:
Isolation is almost impossible, and I see no benefits to them.
Types now definetely work better, I really do not know why you would say they make things worse.
There are no Type-related changes, just refactors.
It was just as bad as it is now, only now its less complicated and you can see that it is wrong.

As to why the unit tests fail: I dont know, they work for me.
Travis apperantly didn't like extracting the listeners to a seperate method, not sure how to go about this.

wandernauta:
Ah, the problem seems to be that it wants to use the `setError(boolean)` version, not the one with three arguments. Should be easy to fix.

i3anaan:
So, how to go further with this?
I see 2 solutions:
- This gets merged in (if no further feedback) its current state. Later on the Type system usage gets overhauled.
- This does not get merged yet, and instead the Type system is overhauled within this PR first.

I think merging now is the way to go.

I am extremely against attempting to cherry pick changes here, as most things are interconnected. I very much disagree with the argument that the Type system is now even worse or harder to improve on. And even if I'm wrong I probably should not be the one cherry picking the Type refactoring out. So if you really want to isolate changes, you may attempt so, I really do not feel like doing so.

jjkester:
My personal opinion is that it would be better to choose the second option (repair types in this PR). I imagine that this will require some major changes, and it would be silly to introduce major changes twice in consecutive pull requests while working on the same thing (this cleanup is done to prepare for type repairs).

Code/src/main/java/nl/utwente/group10/ui/components/blocks/function/FunctionBlock.java
```diff
@@ -0,0 +1,257 @@
...
+        output = new OutputAnchor(this);
+        outputSpace.setCenter(output);
+        
+        // Make sure the prefWidth is correctly updated.
+        this.prefWidthProperty().bind(functionInfo.widthProperty().add(argumentSpace.prefWidthProperty()));
+        
+        
+        this.setMinSize(Region.USE_PREF_SIZE, Region.USE_PREF_SIZE);
+        this.setMaxSize(Region.USE_PREF_SIZE, Region.USE_PREF_SIZE);
+        functionInfo.setMinWidth(Region.USE_PREF_SIZE);
```
TheMSB: @i3anaan  This is where your resizing fails, if you print out the values for prefWidth it gives you the value 60, that is exactly the value that the functionBlock is being sized to.

i3anaan: Sadly It is not.
A simple print statement (As I understand you have used) does not suffice here, since I'm using properties.
When you add the following line you can see the resizing fails somewhere else:

        this.prefWidthProperty().addListener(a -> System.out.println(this.getPrefWidth()));

This gives the following output:
102.0
158.08799743652344
214.17599487304688
270.2639923095703

Meaning that prefWidth is set to 270, which should be about right.

Issue #107, opened 2015-06-09
Context menu polish


i3anaan:
Animation needs to be faster / interruptable
New icons
Some user interactivity

TheMSB:
@i3anaan  Can you specify the problems with the animation in detail? I would like to understand the issue before 'fixing' it.

What do you mean by "Some user interactivity"?

i3anaan:
The animation is currently somewhat slow, but more importantly, uninterruptable.
When the animation starts and you try to pan, the pane instantly jumps to a position (The pane pans in such a way that the context menu jumps to your mouse). This is annoying.
On top of that am I getting the feeling that it opens on things it should not open on (anchors), not too sure about that. Might be that it opens to fast sometimes, hard to explain, mostly just feels off.

What I ment with user interactivity is showing that the buttons are buttons, for example by becoming brighter when you hover above them, and possibly also give visual feedback when clicked.

TheMSB:
So I fixed the animation speed, as for your other points:

- Keep in mind we are using a touch interface, there is no hover over state. I have no idea how you would want me to implement an icon light up or something when I 'hover' over the button.

- If we are using touch visual feedback in general for a button seems like a bad idea, when I press something my finger is on top of it, I won't see any of the stuff happen. Most modern 'flat' UI styles do not feature any such thing either.

PR #121, opened 2015-06-27
ctx Menu fixes

TheMSB:
This makes the animation for the context menu near instantaneous.
Pending my review of what has been changed for context menus in PR #120 I will add more of the requested fixes.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
@@ -1,12 +1,12 @@
 package nl.utwente.group10.ui.menu;
 
-import nl.utwente.group10.ui.components.blocks.Block;
+import javafx.util.Duration;
 
+import nl.utwente.group10.ui.components.blocks.Block;
 import javafx.scene.control.MenuItem;
 import javafx.scene.layout.StackPane;
 import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
-
 import jfxtras.scene.layout.CircularPane;
 import jfxtras.scene.menu.CirclePopupMenu;
 
@@ -57,6 +57,7 @@ public class CircleMenu extends CirclePopupMenu {
 
         // Animation
         this.setAnimationInterpolation(CircularPane::animateOverTheArcWithFade);
+        this.setAnimationDuration(Duration.ONE);
     }
 
     /** Copy the {@link Block} in this context. */


TheMSB:
I looked through the changes and what was left of the issue #107 and came to the conclusion that this is all we would need. Feel free to merge if you believe the fix here is good enough.

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
```diff
@@ -1,12 +1,12 @@
 package nl.utwente.group10.ui.menu;
 
-import nl.utwente.group10.ui.components.blocks.Block;
+import javafx.util.Duration;
 
+import nl.utwente.group10.ui.components.blocks.Block;
```
wandernauta: Do you need this import?

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
```diff
@@ -57,6 +57,7 @@ public CircleMenu(Block block) {
 
         // Animation
         this.setAnimationInterpolation(CircularPane::animateOverTheArcWithFade);
+        this.setAnimationDuration(Duration.ONE);
```
wandernauta: Does setting the interpolation make sense if the animation is only 1ms long?

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
TheMSB: > Do you need this import?

You mean the JavaFX duration? Yes I use it in line 60 the setAnimationDuration only accepts Duration values.

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
TheMSB: > Does setting the interpolation make sense if the animation is only 1ms long?

Yes because without setting it the menu will not show at all, a quirky behaviour from the library.

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
wandernauta: No, the import for nl.utwente.group10.ui.components.blocks.Block.

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
wandernauta: Heh, that's silly.

Code/src/main/java/nl/utwente/group10/ui/menu/CircleMenu.java
TheMSB: Yes the Context menu keeps track of a Block object that serves as it's context.

wandernauta: Ah, you're right, I misread the diff. Thanks a lot! Will merge.

PR #122, opened 2015-06-27
HM: Union typeclasses instead of intersecting them

wandernauta:
I distinctly remember claiming that, when one unifies two variable types
that both have typeclasses, one should take the intersection of their
respective constraints.

This is, of course, patent nonsense. If you have two functions:

    f :: (Show a, Fruit a) => a -> a
    f x = x

    g :: (Show b, Vehicle b) => b -> b
    g y = y

And you define a third function:

    h z = (f z, g z)

Then z must be Show, Fruit *and* Vehicle (a banana mobile, perhaps?)
GHCi will gladly give you the actual signature:

    h :: (Show t, Vehicle t, Fruit t) => t -> (t, t)

In other words, the sets of constraints are unioned, not intersected,
with one another. This commit is a first attempt at that.


jjkester:
That is indeed a bad bug. Haskell hurts my brain on this point, but the example makes it better understandable.

jjkester:
Looks great! Thanks for spotting this :)

PR #125, opened 2015-07-02
Replace mentions of 'String' in catalog with '[Char]'.

wandernauta:
String is a type alias for `[Char]`, but our type checker doesn't know that.
This replaces all mentions of String with `[Char]`.


jjkester:
We should mention this in the user documentation.

PR #128, opened 2015-07-05
Add function definition blocks

wandernauta:
These blocks allow the creation of named lambda's (functions).


Code/src/main/java/nl/utwente/group10/haskell/expr/Function.java
```diff
@@ -60,8 +60,12 @@ public String toString() {
      * @param expr The expression that forms the base of this function.
      */
     public Function(Expr expr, FunctionArgument ... arguments) {
-        this.arguments = Arrays.asList(arguments);
+        this(expr, Arrays.asList(arguments));
+    }
+
+    public Function(Expr expr, List<FunctionArgument> arguments) {
```
jjkester: Could you copy the javadoc to this constructor as well? That way it will be consistent throughout the back end.

wandernauta:
Test failure seems to be caused by Travis having a slightly outdated version of JavaFX which doesn't come with input dialogs.

Code/src/main/java/nl/utwente/group10/ui/menu/FunctionMenu.java
```diff
@@ -143,6 +147,25 @@ private void addFunctionBlock(FunctionEntry entry) {
         addBlock(fb);
     }
 
+    private void addDefinitionBlock() {
```
jjkester: Missing javadoc

Code/src/main/java/nl/utwente/group10/ui/components/blocks/DefinitionBlock.java
```diff
@@ -0,0 +1,128 @@
+package nl.utwente.group10.ui.components.blocks;
```
jjkester: This whole file misses javadoc.

jjkester:
Can we fix Travis? If so, please do.

Code seems fine, but please add the missing javadoc. Good job! :+1:

wandernauta:
Thanks for taking a look. I've added some Javadoc.

There are a few ways to fix Travis:

- Wait until Travis updates their Java 8 JRE to 8u40.
- Force a Java update ourselves.
- Write our own input dialog. (We already have our own message dialog in `CustomAlert`).
- Find some library that does input dialogs and port our stuff to that library.
- Use some sort of combination of Swing and JavaFX that would impress Dr. Frankenstein.

...but neither is quick or painless, I'm afraid.

wandernauta:
Turns out option 2 is relatively painless, see https://github.com/travis-ci/travis-ci/issues/3259. Trying that now.

jjkester:
I think that the first option should be OK, maybe drop Travis a line?

I'll merge, thanks!

PR #130, opened 2015-07-05
Have Travis install newest Java

wandernauta:
This has Travis install Java minor revision 45, which has input dialogs in JavaFX.


wandernauta:
Test failure is weird, not sure if related.

jjkester:
That is something from #128.

PR #136, opened 2015-07-05
Add Hindley-Milner example image

wandernauta:


jjkester:
I do not really like the image, the style does not match anything. Is this fixable (for example text in same font, etc.)?

wandernauta:
I believe I still have the SVG file.

wandernauta:
Yep, I still have the SVG.

wandernauta:
I have the SVG, but changing the font makes Inkscape panic over the use of Greek. I'll close this for now.

PR #146, opened 2015-07-06
Moved Menus and DefinitionBlock

i3anaan:
I see no reason why menu's are not components, so I moved them under components.


TheMSB:
> I see no reason why menu's are not components, so I moved them under components.

Reasoning behind this was that Components are the objects that represent Haskell code, menus don't represent that.

PR #148, opened 2015-07-06
Report remarks

i3anaan:
This is currently a work in progress, but feedback is already welcome.

This will probably change stuff in the entire report, since I am just reading it from top to bottom.


i3anaan:
Done going through the report, turned out a little bigger than expected, although most changes are in the UI sections

PR #152, opened 2015-07-10
Add license, improve README

jjkester:


jjkester:
Feel free to delete the branch once merged.

wandernauta:
Looks great, thanks!

Do you think we have to/should mention TactileAPI/TactileFX somewhere? It's quite a large part of the code base, and it isn't really ours.

jjkester:
Yes, that is a bit of a problem. It really would be nice if TactileAPI becomes a Maven arctifact, but I do not think that will happen soon (or at all)...

What kind of 'mention' do you propose, just a line at the bottom or an explanation about what it is?

wandernauta:
You could go with something like this at the bottom of the license file... 

> Viskell uses code from the TactileAPI project (github.com/...) by permission. 
> TactileAPI is Copyright (c) 2014-2015 TactileAPI authors

...but there is really no way to do this the Right Way as long as TactileAPI doesn't have an official license. Really, any thank-you notice will do.

jjkester:
Updated.

wandernauta:
Fantastic! Will merge on green.

jjkester:
You really think Travis will fail on that last commit ;-)

wandernauta:
...not really, but it's the principle that counts, isn't it? :wink:

PR #159, opened 2015-10-01
Small evaluation improvements

comonoidial:
Fixing a value block issue by adding extra parentheses, just to try github things out.
Also tweak ghci evaluation of displayblocks.
Then chase all these parentheses into the tests...
Removed some dead code along the way.
Concluded that editing code through the github website is not fun.


wandernauta:
Looks good! Will merge, thanks.

PR #169, opened 2015-10-03
Typecheck cleanup

comonoidial:
Fixed the pretty printing of types with repect to parentheses and type variables.
Removed uselss GenSet and moved the typechecker to type package.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
@@ -63,8 +63,9 @@ public class ConstT extends Type {
             out.append(arg.toHaskellType(10));
         }
 
-        if (fixity > 9 && this.args.length > 0)
+        if (fixity > 9 && this.args.length > 0) {
         	return "(" + out.toString() + ")";
+        }
         			
         return out.toString();
     }
--- a/Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
@@ -54,15 +54,18 @@ public class ConstT extends Type {
     }
 
     @Override
-    public String toHaskellType() {
+    public String toHaskellType(final int fixity) {
         StringBuilder out = new StringBuilder();
         out.append(this.constructor);
 
         for (Type arg : this.args) {
             out.append(" ");
-            out.append(arg.toHaskellType());
+            out.append(arg.toHaskellType(10));
         }
 
+        if (fixity > 9 && this.args.length > 0)
+        	return "(" + out.toString() + ")";
+        			
         return out.toString();
     }
 
--- a/Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
@@ -30,7 +30,15 @@ public abstract class Type extends HaskellObject implements Comparable<Type> {
     /**
      * @return The Haskell (type) representation of this type.
      */
-    public abstract String toHaskellType();
+    public String toHaskellType() {
+    	return this.toHaskellType(0);
+    }
+
+    /**
+     * @param The fixity of the context the type is shown in.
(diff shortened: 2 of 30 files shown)


Code/src/main/java/nl/utwente/group10/haskell/type/Type.java
```diff
@@ -30,7 +30,15 @@
     /**
      * @return The Haskell (type) representation of this type.
      */
-    public abstract String toHaskellType();
+    public String toHaskellType() {
+    	return this.toHaskellType(0);
+    }
+
+    /**
+     * @param The fixity of the context the type is shown in.
```
wandernauta: Could perhaps use a short definition, as well as a reference of what the values mean (i.e. that they should always be between 0 and 9).

Code/src/main/java/nl/utwente/group10/haskell/type/ConstT.java
```diff
@@ -54,15 +54,18 @@ public final Type prune() {
...
         out.append(this.constructor);
 
         for (Type arg : this.args) {
             out.append(" ");
-            out.append(arg.toHaskellType());
+            out.append(arg.toHaskellType(10));
         }
 
+        if (fixity > 9 && this.args.length > 0)
+        	return "(" + out.toString() + ")";
```
wandernauta: Code style elsewhere is braces `{}` even for one-liners.

Code/src/main/java/nl/utwente/group10/haskell/type/TypeChecker.java
```diff
@@ -7,34 +7,30 @@
...
 import nl.utwente.group10.haskell.exceptions.HaskellTypeError;
 import nl.utwente.group10.haskell.expr.Expr;
-import nl.utwente.group10.haskell.type.ConstT;
-import nl.utwente.group10.haskell.type.Type;
-import nl.utwente.group10.haskell.type.TypeClass;
-import nl.utwente.group10.haskell.type.VarT;
 
 /**
- * Implementation of the Hindley-Milner type system for Haskell types and expressions.
+ * Implementation of a typechecker for a simplified variant of Haskell.
```
wandernauta: Brutal, but honest. :+1:

Code/src/main/java/nl/utwente/group10/haskell/type/TypeChecker.java
```diff
@@ -7,34 +7,30 @@
...
      * Logger for this class.
      */
-    private static final Logger logger = Logger.getLogger(HindleyMilner.class.getName());
+    private static final Logger logger = Logger.getLogger(TypeChecker.class.getName());
 
     /**
      * Offset for the creation of variable types.
      */
-    static int tvOffset = 0;
+    static int tvOffset = -1;
```
wandernauta: Why `-1`?

Code/src/main/java/nl/utwente/group10/haskell/type/VarT.java
```diff
@@ -132,12 +148,12 @@ public VarT getFresh() {
             instance = this.instance.get().getFresh();
         }
 
-        return new VarT(this.name, this.constraints, instance);
+        return new VarT(this.prefix, this.uid, this.constraints, instance);
```
wandernauta: If you return a new VarT with the same uid, is that uid still a unique identifier?

wandernauta:
Other than these little things, this lgtm :+1: The pretty printing really does look much nicer and less alien this way. Thanks! :sparkles:

comonoidial:
The concept of unique identifier is indeed broken, and is unfornately not easy to fix.
The problem is in displaying type variables to the user, and this depends how define the scope in which they need to be unique in the UI.
It deserves an issue report on its own to define scope in the UI and how that could be implemented.
This pull request will make type variables a little less broken, and making typevars more unique will break unit tests.

wandernauta:
Understood, I'll just go ahead and merge this PR as is. Thanks again!

wandernauta:
Merged in 1d5da88601f08375152a48261268196506eefe16.

Issue #155, opened 2015-09-21
Save/load functionality


Flumble:
Viskell definitely requires a way to save and load Viskell programs. The load and save events are available and a basic load/save system is available in the https://github.com/wandernauta/viskell/tree/saveload branch.

andrewdavidmackenzie:
FYI I'm working on this within the current structure.

andrewdavidmackenzie:
I have done a bit of work on this, with some minor changes to the output format and implementing much of the import functionality.

I have it working for a subset of types of blocks - but some others have some complex sub-elements (FunctionInfo, FunctionReference) that I don't understand yet and will need help to serialize and deserialize.

Here are some of my considerations implementing it, and I will share a WIP PR for you to critique.

# Output Formats
I don't see anything wrong with JSON, it's well known, easy to parse and read.

I thought about using GSON "totally" for "automatic" serialization and deserialization.
One problem is that the JSON output would be full of fully qualified classnames and not so nice and clean and easy to read.

That makes it also fragile to package or class name changes. That will be true (at least with class names or object names) for almost any solution.

# Serialization
I've gone the route of adding more data to the export, splitting it into two collections (JSON Arrays) of two different types ("Blocks" and "Connections"). I was going to serialize into three layers as is structured in memory, but saw that addBlock() figures out which layer a block belongs to and adds accordingly. So I can serialize all blocks into one array and deserialization just adds them all using addBlock().

# Deserialization
When I looked more at the code and the need to connect things up after deserialization I decided that deserialization with the help of all the involved objects (like serialization) was probably the best way to go.

There are two main ways I considered to do this:
1) All serializable classes have a default (zero parameter constructor). This is instantiated by the deserializer, then an instance method is called, passing in the Bundle and it sets it own class-specific parameters from the Bundle.

2) All seriablizable classes have a static fromBundleFragment() method that reads in a bundle, and instantiates an object using the existing constructors, reading extra class-specific parameters from the bundle and returns the new object.

I decided on 1) as it was less disruptive, avoiding the need for new Constructors for all classes AND the fromBundle() method - and keeping those constructors in-sync with existing ones. Static method maybe is not so nice, but it's less code and any serialization related changes and limited to this method. It also uses the existing Constructors (some of them have non-trivial logic....)

# Deserialization Process
The input file is read and GSON is used to parse into two arrays, one for Blocks and one for Connections.

[...]

PR #172, opened 2015-10-03
Add a global menu for global actions

wandernauta:
Adds a hamburger button to the zoom overlay (which is now called the
button overlay) that shows a dropdown (context) menu when clicked.

The menu only has a Quit action at present, but this is where the New, Open,
Save and Save As actions from #155 would go.


comonoidial:
Quick note, it might be handy to add the inspector and debug mode to this menu too.

comonoidial:
Quick note, the current implementation seems to block all mouse actions across the entire top row of the screen for about 40 pixels.

wandernauta:
@comonoidial Yep. There's two ways to fix this:

- Give the top bar some kind of background to indicate that there's going to be more buttons on there (there aren't many yet, but there might be).
- Set the top bar mouse transparent so that it no longer catches mouse events.

comonoidial:
Setting the bar mouse transparent also disables the button, and I don't like reducing screen space.
So I guess these flowpanes should be replaces by gridpanes instead.

wandernauta:
I'm not sure gridpane fits. I thought there was a way to tell flowpane to shrink to its contents; probably some magic incantation with `minWidth`, `prefWidth`, `maxWidth` etc

PR #175, opened 2015-10-03
Fix TypeVar instance bug

comonoidial:
Add missing symmetric case in de typechecker unification.
Enforce correct instancing behaviour in TypeVar.
And some debugging improvements

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/haskell/type/FunType.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/type/FunType.java
@@ -61,10 +61,6 @@ public class FunType extends ConcreteType {
 
     @Override
     public String toString() {
-        final StringBuilder out = new StringBuilder();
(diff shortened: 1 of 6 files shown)


Code/src/main/java/nl/utwente/group10/haskell/type/FunType.java
```diff
@@ -61,6 +61,10 @@ public boolean containsOccurenceOf(TypeVar tvar) {
 
     @Override
     public String toString() {
-        return this.toHaskellType();
+        final StringBuilder out = new StringBuilder();
+        out.append(this.argument.toString());
+        out.append(" -> ");
+        out.append(this.result.toString());
+        return out.toString();
```
wandernauta: Why not `String.format("%s -> %s", argument, result)`?

wandernauta:
Thanks!

Issue #177, opened 2015-10-04
Add support for type application in the type parser


comonoidial:
After PR #176 it looks like the parser is the only thing blocking the addition of functors and alike to the catalog.
The parser/type builder should be extended to support types like:
fmap :: Functor f => (a -> b) -> f a -> f b
sequence_ :: Monad m => [m a] -> m ()
traverse :: (Traversable t, Applicative f) => (a -> f b) -> t a -> f (t b)
first :: Arrow a => a b c -> a (b, d) (c, d)

wandernauta:
I've linked a branch that adds rudimentary type application parsing to the type parser.

I think we also need something else, though - Viskell must be able to deduce that a list (any list) is a Functor, for example. I'm not yet seeing how that would fit in the catalog.

comonoidial:
I think something along the following lines will do just fine:
```
       <class name="Functor">
        	<instance name="Maybe"/>
        	<instance name="[]"/>
        	<instance name="IO"/>
        </class>
```

wandernauta:
Right, that looks reasonable. I'll give it a try.

wandernauta:
Sweet, seems to work for Maybe:

![2015-10-04-163558_653x616_scrot](https://cloud.githubusercontent.com/assets/126832/10268348/0707495a-6ab6-11e5-98b5-9208123ea1c7.png)

`[]` is not a type constructor according to the parser, because it doesn't start with a capital letter. I'll try to correct that.

wandernauta:
See #178.

comonoidial:
Fixed by #178

PR #178, opened 2015-10-04
Add type application in the type parser (#177)

wandernauta:
This parses every combination of a type variable followed by something else
as a type application.

Also adds the Functor type class and the fmap function to the catalog.

### DIFF
--- a/Code/src/test/java/nl/utwente/group10/haskell/typeparser/TypeBuilderTest.java
+++ b/Code/src/test/java/nl/utwente/group10/haskell/typeparser/TypeBuilderTest.java
@@ -34,11 +34,13 @@ public class TypeBuilderTest {
         Map<String, TypeClass> typeClasses = new HashMap<>();
         typeClasses.put("Num", new TypeClass("Num", Type.con("Int"), Type.con("Float"), Type.con("Double")));
         typeClasses.put("Eq", new TypeClass("Eq", Type.con("Int"), Type.con("Float"), Type.con("Double"), Type.con("Char"), Type.con("Bool")));
+        typeClasses.put("Functor", new TypeClass("Functor"));
         TypeBuilder builder = new TypeBuilder(typeClasses);
 
         Assert.assertEquals("(Num a)", builder.build("Num a => a").toHaskellType());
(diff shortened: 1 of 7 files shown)


comonoidial:
Also for '[]'  special case is needed somewhere in the type building process to ensure it ends up as an ListTypeCon

wandernauta:
Is the arity of a tuple contstructor the number of arguments or the number of comma's? Those seem to be off by one.

comonoidial:
Arity is the term refering to the number of arguments.
And the current TupleCon code is inconsistent with that.

wandernauta:
The type builder now uses tuple and list constructors.

comonoidial:
Roundtrips should be added to TypeBuilderTest for at least: () ,  (,)  , (,,) and [].

Code/src/test/java/nl/utwente/group10/haskell/typeparser/TypeBuilderTest.java
```diff
@@ -29,16 +29,22 @@ private void roundtrip(String hs) {
     @Test public void testMaybeMaybe()  { this.roundtrip("Maybe (Maybe a)"); }
     @Test public void testMaybeSink()   { this.roundtrip("Maybe Int -> Maybe [a]"); }
     @Test public void testKitchenSink() { this.roundtrip("[a] -> [b] -> [(a, b)]"); }
+    @Test public void testPrefixUnit()  { this.roundtrip("()"); }
+    @Test public void testPrefixTuple() { this.roundtrip("(,) a"); }
+    @Test public void testPrefixTriple(){ this.roundtrip("(,,) a b"); }
+    @Test public void testPrefixList()  { this.roundtrip("[] a"); }
```
comonoidial: Why does roundtrip("[] a") succeed?
I would expect that "[] a" comes out as the special cased haskell type "[a]"

comonoidial:
Looks all good to me :+1:
It's very nice that most of the standard libraries could used now without reimplementing type classes.

wandernauta:
Merged, thanks for the hints @comonoidial.

PR #179, opened 2015-10-04
Refactoring and cleanup of catalog and environment

comonoidial:
Gets rids of lots of convoluted intermediate data structures and classes.

PR #180, opened 2015-10-05
Implementation of superclass constraint reduction

comonoidial:
This depends on PR #179
Now constraints like (Ord a, Eq a) and (Num n, Integral n) will simplify to (Ord a) and (Integral n), thus keeping types a lot more readable.


wandernauta:
Do we still need the duplicate instances in `catalog.xml`? For example, `Int` is in `Real` as wel as `Num`.

comonoidial:
Technically we could do without it, but that will make it easier to overlook instance.
Now filling in the catalog correctly is basicly copying what come out of ghci :i SomeClass or :i SomeConstructor.

PR #184, opened 2015-10-05
Expression package refactoring

comonoidial:
A lot of code has changed, but most of it is cleaning up things.
The main real changes are in variable expressions, in the use of functions from the environment and the addition of 'Holes' for open inputs.


Code/src/main/java/nl/utwente/group10/ghcj/GhciSession.java
```diff
@@ -55,6 +59,33 @@ public String pull(final Expr expr) throws HaskellException {
...
+     * Ask ghci for the type of an expression
+     * @param expr The expression String to determine the type of.
+     * @param env The environment in which the type will be resolved 
+     * @return The parsed Haskell type
+     * @throws HaskellException when  ghci encountered an error or the type could not be parsed.
+     */
+    public Type pullType(final String expr, Environment env) throws HaskellException {
+        String[] parts = this.pullRaw(":t (" + expr + ")").split(" :: ");
+        if (parts.length < 2) {
+            throw new HaskellException("ghci could not determine the type of:/n" + expr);
```
wandernauta: Should probably be `\n` instead of `/n`.

Code/src/main/java/nl/utwente/group10/ghcj/GhciSession.java
```diff
@@ -55,6 +59,33 @@ public String pull(final Expr expr) throws HaskellException {
...
+    
+    /**
+     * Ask ghci for the type of an expression
+     * @param expr The expression String to determine the type of.
+     * @param env The environment in which the type will be resolved 
+     * @return The parsed Haskell type
+     * @throws HaskellException when  ghci encountered an error or the type could not be parsed.
+     */
+    public Type pullType(final String expr, Environment env) throws HaskellException {
+        String[] parts = this.pullRaw(":t (" + expr + ")").split(" :: ");
```
wandernauta: There's an extra pair of parentheses around the expression here - do we also need that in push, above?

Code/src/main/java/nl/utwente/group10/haskell/env/Environment.java
```diff
@@ -0,0 +1,93 @@
...
+public class Environment {
+    /**
+     * Map containing the types of function available in the environment.
+     */
+    private Map<String, FunctionInfo> functions;
+
+    /**
+     * Map containing the type classes by name.
+     */
+    private HashMap<String, TypeClass> typeClasses;
```
wandernauta: Might be able to weaken this to `Map<>`.

Code/src/main/java/nl/utwente/group10/haskell/env/Environment.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * @param type The String representation of the type signature
+     * @return Type built using the context of this environment
+     */
+    public final Type buildType(String type) {
+        return new TypeBuilder(this.typeClasses).build(type);
+    }
+
+    /**
+     *
```
wandernauta: Missing Javadoc.

Code/src/main/java/nl/utwente/group10/haskell/env/Environment.java
```diff
@@ -0,0 +1,93 @@
...
+    public final TypeClass lookupClass(String name) {
+        return this.typeClasses.get(name);
+    }
+    
+    /**
+     * Adds an type signature for function (lacking implementation) to this environment.
+     * @param name The name of the expression.
+     * @param signature The signature of the expression.
+     */
+    public final void addTestSignature(String name, String signature) {
```
wandernauta: Maybe make an `Environment` subclass that allows this, and use that subclass from tests?

Code/src/main/java/nl/utwente/group10/haskell/expr/Apply.java
```diff
@@ -29,25 +28,22 @@
...
+        final Type argType = arg.inferType();
         final Type resType = TypeChecker.makeVariable("b");
 
-
         // Rule [App]:
         // IFF  the type of our function is a -> b and the type of our arg is a
         // THEN the type of our result is b
         TypeChecker.unify(this, funcType, Type.fun(argType, resType));
 
-        this.setCachedType(resType);
```
wandernauta: Why can this result not be cached?

Code/src/main/java/nl/utwente/group10/ui/InspectorWindow.java
```diff
@@ -75,11 +75,11 @@ public void update() {
     /**
      * Walks the expr tree, walk recursively calls itself on its children.
      */
-    private void walk(TreeItem<String> treeItem, Expr expr) {
+    private void walk(TreeItem<String> treeItem, Expression expr) {
         String type;
 
         try {
-            type = expr.getType(pane.getEnvInstance()).toHaskellType();
+            type = expr.findType().toHaskellType();
```
wandernauta: Does `findType` change the expression in any way/do any inference? If so, this change might cause additional confusion (in the category 'it only works if I have the Inspector open').

wandernauta:
Other than the noted nitpicks, this looks great :+1:

comonoidial:
The caching logic has moved to a central place.
I renamed getType to findType because it's not a simple getter.
Some of the early tests use hacky ways to do things, I would prefer refactoring these tests over adding special testing classes.
And the rest are trivial oversights on my part.

PR #186, opened 2015-10-05
Added support for lambda expression with type inference

### LINKED ISSUE #153: Add support for inferring the type of lambdas
In the current implementation, the input and output types of lambdas (definition blocks) have to be given at creation time.

It would be preferable if Viskell were able to infer the types of lambdas in the same way it infers the types of other expressions.

comonoidial:
Relatively small change but strongly depends on PR #184.
Except for the UI integration part this resolve #153.


Code/src/main/java/nl/utwente/group10/haskell/expr/Binder.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public final class Binder {
+
+    /** The name of this binder */
+    private final String name;
+    
+    /** An internal type used in the type inference process */
+    private TypeVar inferenceType;
+    
+    /** An optional */
```
wandernauta: An optional what?

Code/src/main/java/nl/utwente/group10/haskell/expr/Binder.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @return The base name of this binder
+     */
+    public String getBaseName() {
+        return this.name;
+    }
+    
+    /**
+     * @return the type annotation, might be null
```
wandernauta: Prefer `Optional<T>` over hard nulls.

Code/src/main/java/nl/utwente/group10/haskell/expr/Binder.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.util.IdentityHashMap;
+
+import nl.utwente.group10.haskell.exceptions.HaskellTypeError;
+import nl.utwente.group10.haskell.type.Type;
+import nl.utwente.group10.haskell.type.TypeChecker;
+import nl.utwente.group10.haskell.type.TypeVar;
+
+/**
+ * A Binder is the definition side of a local variable, it is used link variable to their binding constructs, such as lambdas
```
wandernauta: Do you have an example of a Binder?

Code/src/main/java/nl/utwente/group10/haskell/expr/Lambda.java
```diff
@@ -0,0 +1,77 @@
...
+        StringBuilder out = new StringBuilder();
+        out.append("(\\");
+
+        for (Binder x : this.binders) {
+            out.append(" ").append(x.getUniqueName());
+        }
+
+        out.append(" -> ");
+        out.append(this.body.toHaskell());
+        out.append("");
```
wandernauta: Appends an empty string?

wandernauta:
Neat! And tested, too! :sparkles:

PR #189, opened 2015-10-07
Fix for two tricky typechecker bugs

comonoidial:
As this branch was first intended for something else it also contains a few cleanups and preparational refactorings.


Code/src/main/java/nl/utwente/group10/haskell/type/TypeClass.java
```diff
@@ -1,32 +1,76 @@
...
+     */
+    final static class Instance implements Comparable<Instance> {
+
+        /**
+         * The type constructor that identifies this instance.
+         */
+        private final TypeCon typecon;
+        
+        /**
+         * The number of type arguments that needs to be constrained to make this instance valid
```
wandernauta: I like this Javadoc.

Code/src/main/java/nl/utwente/group10/haskell/type/TypeScope.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.TreeSet;
+
+public class TypeScope {
+    /**
+     * Offset for the creation of unique type variable names.
+     */
+    static int tvOffset = 0;
```
wandernauta: Should this be static?

Code/src/main/resources/catalog/catalog.xml
```diff
@@ -9,15 +9,35 @@
             <instance name="Double"/>
             <instance name="Char"/>
             <instance name="Bool"/>
+            <instance name="[]" constrainedArgs="1"/>
```
wandernauta: Maybe it would be nicer to have the instance name as "[a]" or "[] a" or "Either a b" here? These instance names go through the type parser anyway, and that way you wouldn't need the constrainedArgs counter.

Code/src/main/java/nl/utwente/group10/haskell/type/TypeVar.java
```diff
@@ -33,17 +31,26 @@
...
          */
-        private Set<TypeClass> constraints;
+        private TreeSet<TypeClass> constraints;
+        
+        /**
+         * Internal List of all type variables that refer to this instance.
+         * this list contains the creating type variable as well as the variables have unified with this.
+         * It is using WeakReference to avoid holding on to too much old type variables. 
+         */
+        private LinkedList<WeakReference<TypeVar>> unifiedVars;
```
wandernauta: What is in here, exactly? And when does it leave the list?

PR #193, opened 2015-10-12
Add support for simple type classes on nested types

comonoidial:
Show, Ord, and alike on lists, tuples, Maybe and more.
depends on PR #191

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/haskell/type/TypeClass.java
+++ b/Code/src/main/java/nl/utwente/group10/haskell/type/TypeClass.java
@@ -88,8 +88,8 @@ public class TypeClass implements Comparable<TypeClass> {
     /**
      * @return the super classes of this type class
      */
-    public Set<TypeClass> getSupers() {
-        return supers;
+    protected Set<TypeClass> getSupers() {
+        return this.supers;
     }
     
     /**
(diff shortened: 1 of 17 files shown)


Code/src/main/java/nl/utwente/group10/haskell/type/TypeClass.java
```diff
@@ -89,6 +86,13 @@ public final void addInstance(TypeCon tc, int constrainedArgs) {
     }
 
     /**
+     * @return the super classes of this type class
+     */
+    public Set<TypeClass> getSupers() {
+        return supers;
```
wandernauta: Is the caller of this method allowed to modify the set it gets afterwards?

Code/src/main/java/nl/utwente/group10/haskell/type/TypeVar.java
```diff
@@ -207,35 +189,36 @@ public final void setConcreteInstance(ConcreteType type) {
...
     }
 
     /**
-     * @return Whether this type variable is constrained.
+     * @return The set of type class constraints associated with this type variable
      */
-    public final boolean hasConstraints() {
-        return !this.instance.constraints.isEmpty();
+    ConstraintSet getConstraints() {
+        return this.instance.constraints;
```
wandernauta: How about this method? Can the caller of this method introduce extra constraints on the result it gets?

wandernauta:
Other than some OO nitpicks this looks marvelous :+1:

Issue #185, opened 2015-10-05
Super issue: the relations between the UI structure, the types and Haskell expressions


comonoidial:
Most of the issues in viskell are caused by how the UI structure relates to the generated haskell expression, and how type construction and typechecking interacts with the UI.
While some of things could be solved by involved workaround, a thorough redesign of these aspect is required.
The changes will be invasive and needs discussion first, but it has the highest priority once the current code has been cleaned up.

Without addressing the core design flaws, solving issues  #161, #162, #170 and #183 will be difficult.
And also issues #154, #155, #158, #166, #168 and #171 would all benefit from a good solution.

comonoidial:
This issue is too broad with no clear solution in sight, resolution defered to specific sub issues.

PR #195, opened 2015-10-13
Add save/save as actions and JSON serializer

wandernauta:
This PR implements the 'Save' and 'Save as' actions. 

The (JavaFX) components are not Java `Serializable`. Instead of using Java's built-in serialization, this PR serializes the program graph to a nicely formatted JSON tree. Components can mark themselves as serializable by implementing the `Bundleable` interface.

The JSON serialization is shown in the Inspector to help with debugging.

### DIFF
--- a/Code/pom.xml
+++ b/Code/pom.xml
@@ -19,6 +19,11 @@
       <version>18.0</version>
     </dependency>
     <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.4</version>
+    </dependency>
+    <dependency>
       <groupId>org.antlr</groupId>
       <artifactId>antlr4-runtime</artifactId>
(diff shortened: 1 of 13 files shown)


comonoidial:
I don't see anything strange in this code, although I have no clue what the gson library is doing.
When loading is added eventual problems will show themselves.
And I wonder how this code will might need to be adapted to a resolution of #185.
It doesn't affect any other component, so imho just ship it if it compiles.

wandernauta:
The gson library here does the step from 'tree of maps' to 'JSON-formatted string'. This step is admittedly almost trivial; with JSON as with most languages, generating is easier than parsing. We don't do parsing yet, but when we do, gson can help us there, as well.

A good solution to #185 should make loading easier - it's definitely the difficult part of the equation right now because 'program parts' are so tightly coupled with how they look/their JavaFX representations.

Issue #194, opened 2015-10-12
Function application deserves it own kind of block


comonoidial:
Currently it is easy to create blocks that have a function as an output, by currying or definition blocks.
But when an input is a function using it is inconvenient and one has to use combinators like ($).
For example see what is needed now to implement something alike to the standard library function:
on :: (b -> b -> c) -> (a -> b) -> (a -> a -> c)

Using functions is as important as creating them, so a special function block would be nice to have.
It could be like a function block that takes a function wire instead of a name, or maybe someone has a better idea.

wandernauta:
Agreed - it could even be something as simple as this:

![mockup](https://cloud.githubusercontent.com/assets/126832/10455026/02f0e5f6-71b9-11e5-9096-f056a696cbda.png)

That would compile to either a use of `($)` or just an application, I'm not sure if there's any difference.

comonoidial:
I would prefer something more specialized.
Having the function type as first element is superfluous information.
Also the bowtie thing makes no sense for single argument functions, and currently behaves oddly on ($) with multi-argument functions.

comonoidial:
Resolved in #374 plus #384.

Issue #158, opened 2015-10-01
Add wire splitters and joiners


comonoidial:
Constructing and matching on tuples is a very common thing to do, and joining/splitting wires seems a very natural metaphor.
I think we should add a special nameless block for both a splitter and a joiner.
As this is a very basic variation on pattern matching, it might be something for @Flumble to start with.
The main complication is that splitters need multiple outputs.

wandernauta:
As a first step, I think a shortcut to the `(,)` function would work well as a (2-way) joiner.

comonoidial:
Resolved by #360.

PR #200, opened 2015-10-18
Attempt to make the UI look a little more subtle

comonoidial:
Just details, mostly to experiment with how javafx works.

### DIFF
--- a/Code/src/main/java/nl/utwente/group10/ui/components/menu/FunctionMenu.java
+++ b/Code/src/main/java/nl/utwente/group10/ui/components/menu/FunctionMenu.java
@@ -192,7 +192,7 @@ public class FunctionMenu extends StackPane implements ComponentLoader {
         parent.getChildren().add(block);
         block.setConnectionState(ConnectionCreationManager.nextConnectionState());
         Point2D pos = this.localToParent(0, 0);
-        block.relocate(pos.getX() - 200, pos.getY());
+        block.relocate(pos.getX() - 250, pos.getY());
     }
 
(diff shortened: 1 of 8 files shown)


wandernauta:
Interesting. Do the rounded borders have a special meaning?

comonoidial:
Not yet, but possibly is block shape nice to use to distinguish between various kinds, and then pure functions/values should have a minimalistic shape.

wandernauta:
Hmm. I've got the same doubts here that I have about #158 and #194. Every visual element or visual distinction we introduce increases the size of Viskell's vocabulary. For example, `ValueBlock` here has rounded corners and reduced padding, but `DisplayBlock` doesn't. You'd have to explain why they look different to every single user, in some way or other.

comonoidial:
In this case it is not the intention to increase the Viskell's vocabulary.
It is part of a larger experiment I want to do for increasing readability of larger programs, where instead of keeping everything uniform, subtle meaningful variations in visual style are used to subconsciously guide you through a program (these variations have no effect on the generated code).

That displayblock's and co look differently is in my opinion beneficial as they are not really part of the program, but merely observers/debugging helpers of the program.

As for the other mentioned issues, Viskell is now extremely minimal from the syntactical point of view, especially compared to textual languages.
And I think while a visual language needs to have a lot less constructs than a textual one, specializing the most common constructs will help readability by avoiding being a big blob of identical blocks and wires.

comonoidial:
This PR is an outdated mix of things on issues that require a more thorough redesign.

PR #201, opened 2015-10-18
Flatten components package

wandernauta:
This puts all components in the same package and moves some classes that aren't strictly components out of `components`.


comonoidial:
Looks fine to me.

Maybe other packages can be flattened too:
haskell.env and ghcjs -> env
haskell.type and haskell.typeparser --> type
haskell.expr --> expr

wandernauta:
I've considered moving the typeparser to the `type` package, but I decided against it - right now, nothing in the `type` package depends on ANTLR, which is a good thing, I think.

PR #202, opened 2015-10-18
Switch definitionblock over to Lambda and remove Function expr

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/components/DefinitionBlock.java
```diff
@@ -26,20 +30,20 @@ public OutputAnchor getOutputAnchor() {
...
+            this.binder = binder;
+            this.expr = new LocalVar(binder);
 
             anchor = new OutputAnchor(this);
         }
 
         /** Returns the FunctionArgument expression we built. */
-        public Function.FunctionArgument getArgument() {
-            return (Function.FunctionArgument) getExpr();
+        protected Binder getBinder() {
```
wandernauta: Is this method's Javadoc still applicable?

PR #203, opened 2015-10-19
First steps in cleaning up the UI code

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -116,51 +118,65 @@ public final CustomUIPane getPane() {
...
+     * @return Only the active (as specified with by the knot index) inputs.
+     */
+    public List<InputAnchor> getActiveInputs() {
+        return getAllInputs();
+    }
+    
+    /**
+     * @return the optional output Anchor for this Block
+     */
+    public Optional<OutputAnchor> getOutputAnchor() {
```
wandernauta: Could consider making this

    public List<OutputAnchor> getOutputAnchors()

if we want to go N inputs/N outputs in the future.

Code/src/main/java/nl/utwente/viskell/ui/components/ConnectionAnchor.java
```diff
@@ -25,6 +32,114 @@
  */
 public abstract class ConnectionAnchor extends StackPane implements ComponentLoader {
 
+    /**
+     * Handler class that reacts to user inputs on ConnectionAnchors to be able to
+     * create, edit and drag Connections.
+     */
+    private class AnchorHandler implements EventHandler<InputEvent> {
```
wandernauta: Does this have to be a nested class? Can't you just pull its methods into the ConnectionAnchor class?

Code/src/main/java/nl/utwente/viskell/ui/components/Connection.java
```diff
@@ -226,7 +226,7 @@ private void checkErrorListener(ObservableValue<? extends Boolean> value, Boolea
     /**
      * @return Whether or not both sides of this Connection are connected to an Anchor.
      */
-    public final boolean isConnected() {
+    public final boolean isFullyConnected() {
```
wandernauta: Why the name change/revert? Is there an `isHalflyConnected`?

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -116,51 +118,65 @@ public final CustomUIPane getPane() {
...
-     * Updates the expression, clearing the expression dirty flag.
+     * Updates the expression
      */
-    public void updateExpr() {
-        setExprIsDirty(false);
-    }
+    public abstract void updateExpr();
     
-    @Override
     public final int getVisualState() {
```
wandernauta: These methods are now undocumented.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -116,51 +118,65 @@ public final CustomUIPane getPane() {
...
      * @return The expression this Block represents.
      * 
      * If the expression is not up-to-date it gets updated.
      */
-    public Expression getExpr() {
+    public final Expression getExpr() {
         // Assure expr is up-to-date.
         if (getExprIsDirty()) {
             updateExpr();
+            setExprIsDirty(false);
```
wandernauta: Is this the only place that's allowed to touch the dirty flag? If so, maybe we can reduce the visibility on that flag.

Code/src/main/java/nl/utwente/viskell/ui/components/Connection.java
comonoidial: isConnected exists already as unrelated internal method in javafx.
And fullyConnected is clearer (I was also trying to track down a bug with partially connected wires staying on the uipane)

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
comonoidial: Seems to be one other place, but cleaning up the whole getter/setter/property/visibility mess is a later step.

Code/src/main/java/nl/utwente/viskell/ui/components/ConnectionAnchor.java
comonoidial: Eventually it should not be an inner class anymore.
But I rather get rid of the ConnectionCreationManager first.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
comonoidial: Will add an todo, instead of making less obvious changes in cleanup.

wandernauta:
Understood, thanks!

PR #204, opened 2015-10-19
Cut down on cruft in ui properties and connections

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -108,6 +93,7 @@ protected void createCircleMenu() {
     private void handleMouseEvent(MouseEvent t) {
         parentPane.setSelectedBlock(this);
         if (t.getButton() == MouseButton.SECONDARY) {
+            CircleMenu circleMenu = new CircleMenu(this);
```
wandernauta: This introduces a regression where right-clicking multiple times on a block creates multiple circle menu's.

Code/src/main/java/nl/utwente/viskell/ui/components/ConnectionAnchor.java
```diff
@@ -171,42 +165,23 @@ public ConnectionAnchor(Block block) {
...
-    public BooleanProperty activeStateProperty() {
-        return activeState;
-    }
-    
-    /**
-     * @return Whether or not this ConnectionAnchor is in an error state.
-     */
-    public boolean getErrorState() {
-        return errorState.get();
+        if (!active) {
```
wandernauta: With the new implementation, this isn't really a setter anymore.

wandernauta:
This looks great! Let me know when you've taken a look at the (minor) remarks above.

comonoidial:
Resolved in the #205 branch :confused:
All these branches, I just want to work in terms of patches...

wandernauta:
Merged, thanks!

If you really want to, you can just send in patches and I'll `git apply` them for you. I'm not sure how that would make things easier, though.

PR #205, opened 2015-10-20
Fix connection creation to avoid leaving partial ones in the ui in some corner cases

comonoidial:
depends on PR #204.


wandernauta:
:+1: Thanks!

Issue #207, opened 2015-10-20
Rethink the update/change propagation strategy


comonoidial:
The change propagation is now quite a hack with two state counters and a dirty flag in each block.
And it will probably not work for more complicated features such as recursion.
I do not know yet what a good solution is, but for changes like type modifications just a global update might not be bad.

comonoidial:
I consider this as resolved by #217 and #221.

PR #208, opened 2015-10-20
Untangle the logic in creating/editing connections

comonoidial:


comonoidial:
This is all the cleanup that can be done without changing functionality.

wandernauta:
Thanks! I'll take a look right away.

wandernauta:
Merged!

Deleting blocks does not cause the rest of the program to be updated correctly. I thought this was a regression, but the same problem occurs in master. To reproduce the problem, attach the output of a `(*)` block to the input of a `enumFrom` block, then delete the `enumFrom` block. `CustomUIPane.removeBlock` should probably touch some dirty flag somewhere (#207).

PR #209, opened 2015-10-21
Use proper removal methods in removeBlock

comonoidial:
Solves updating of connected blocks on block removal as observed in #208.


wandernauta:
Fantastic, thanks!

PR #210, opened 2015-10-22
Add an untyped lambda variation of the definition block

comonoidial:


wandernauta:
Do we still need the typed variation with this change?

comonoidial:
Explicitly typed toplevel functions are very common in Haskell, even though it's often not strictly needed.

PR #214, opened 2015-10-25
Show value in slider and type in display block

comonoidial:
It doesn't look quite nice yet, but had enough of javafx for today.


wandernauta:
Code looks reasonable :+1: Do you want to merge as is or give the type display on the display block a bit more breathing room first?

comonoidial:
I'm fine with merging it now, learning javafx properly will take some more time.
And the UI needs lots of tweaking anyway.

PR #216, opened 2015-10-25
Experiment with lighter background and darker wires

comonoidial:
The dark background felt too distracting to me, so made it lighter.
Maybe the grid could be made even more subtle.


wandernauta:
I'm hesitant to merge this, because I personally don't think it's an improvement. The first versions had a white background. I originally suggested the darker background with the large multi-touch screen in mind - I'd rather not spend my days staring into what is essentially a light table.

However, I do recognize that it's above all a personal preference (my terminals and editors are all light-on-dark).

I would merge this without a grumble if the color scheme was selectable by the user (maybe even `java.util.prefs`-backed).

comonoidial:
This one is too outdated

Issue #188, opened 2015-10-06
Displaying types in the UI seems all over the place


comonoidial:
See the following two examples, and look at the difference between these.
![wierd_types](https://cloud.githubusercontent.com/assets/14928502/10318679/dfe1e892-6c6a-11e5-8f5c-236ebef15b1d.png)
![weird_types2](https://cloud.githubusercontent.com/assets/14928502/10318684/e611ca2a-6c6a-11e5-9a30-3b241c1d0bb1.png)

Flumble:
Fixed in #217

PR #217, opened 2015-10-29
Rewrite the change propagation logic

comonoidial:
This is attempt to make the change update and propagation logic saner and more correct.
While this improve some of the problems in #188, weird bugs still happen within function block types.
Also depends on PR #210


wandernauta:
Can you elaborate a bit on how (why) this works? If I read the diff correctly, `connectionState` is gone, `exprIsDirty` stays, and what used to be `visualState` is now a single observable boolean `staleVisuals` (not `visualsAreStale`). Touching the staleVisuals bit re'draws' components. So far so good.

However, from what I can tell, changes seem to be propagated both 'up' and 'down'. How do you avoid (infinite) loops? Doesn't this cause more redraws than necessary?

comonoidial:
How can this produce merge conflicts?
This just is based on pr210 merged with yesterday's master...

wandernauta:
It should be a breeze to rebase then. If you're not comfortable rebasing I can just manually merge it for you.

comonoidial:
I think issue #188 is fully resolved now, and also made updating and typechecking even less frequent.
rebasing? why can't git take history in account properly when merging...

wandernauta:
Merged.

Issue #215, opened 2015-10-25
Types of value and slider blocks are wrong


comonoidial:
Value blocks seem to remember their old type even after a full disconnect, which makes them not reusable and causing unnecessary type errors.
Slider block now always have the type Float as output, this is a poor default and too restrictive.

wandernauta:
Maybe the slider block could be turned into a block with two inputs (min, max) and one output, then base the output type on the input type.

comonoidial:
While generalizing slider blocks is a good idea, it is not how the typing problem should be solved.
Also by requiring extra inputs the slider block will be less convenient to use, because where do these inputs come from? (another slider block, oops...)

wandernauta:
The inputs wouldn't need to be required - they could default to 0.0 and 1.0 respectively.

comonoidial:
Fixed in #218.

PR #218, opened 2015-10-30
Issue215

Flumble:
This pull request resolves issue #215

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/components/SliderBlock.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/components/SliderBlock.java
@@ -2,7 +2,7 @@ package nl.utwente.viskell.ui.components;
 
 import javafx.fxml.FXML;
 import javafx.scene.control.Slider;
-import nl.utwente.viskell.haskell.type.Type;
+import nl.utwente.viskell.haskell.env.Environment;
 import nl.utwente.viskell.ui.CustomUIPane;
 
 /**
@@ -11,14 +11,14 @@ import nl.utwente.viskell.ui.CustomUIPane;
  * Ranges from 0 to 1 (both inclusive).
  */
 public class SliderBlock extends ValueBlock {
-    @FXML private Slider slider;
+    @FXML protected Slider slider;
 
     /**
      * Constructs a new SliderBlock
      * @param pane The parent pane this Block resides on.
      */
-    public SliderBlock(CustomUIPane pane) {
-        super(pane, Type.con("Float"), "0.0", "SliderBlock");
+    public SliderBlock(CustomUIPane pane, Environment env) {
+        super(pane, env.buildType("Fractional a => a"), "0.0", "SliderBlock");
 
         slider.setValue(0.0);
         this.setValue("0.0");
--- a/Code/src/main/java/nl/utwente/viskell/ui/components/SliderBlock.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/components/SliderBlock.java
@@ -17,8 +17,8 @@ public class SliderBlock extends ValueBlock {
      * Constructs a new SliderBlock
      * @param pane The parent pane this Block resides on.
      */
-    public SliderBlock(CustomUIPane pane, Environment env) {
-        super(pane, env.buildType("Fractional a => a"), "0.0", "SliderBlock");
+    public SliderBlock(CustomUIPane pane) {
+        super(pane, pane.getEnvInstance().buildType("Fractional a => a"), "0.0", "SliderBlock");
 
         slider.setValue(0.0);
         this.setValue("0.0");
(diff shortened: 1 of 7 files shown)


wandernauta:
Thanks, I'll take a look.

Code/src/main/java/nl/utwente/viskell/ui/components/SliderBlock.java
```diff
@@ -11,14 +11,14 @@
...
+    @FXML protected Slider slider;
 
     /**
      * Constructs a new SliderBlock
      * @param pane The parent pane this Block resides on.
      */
-    public SliderBlock(CustomUIPane pane) {
-        super(pane, Type.con("Float"), "0.0", "SliderBlock");
+    public SliderBlock(CustomUIPane pane, Environment env) {
+        super(pane, env.buildType("Fractional a => a"), "0.0", "SliderBlock");
```
wandernauta: Why pass `env` as an argument when you can just as easily do `pane.getEnvInstance` here?

Code/src/main/java/nl/utwente/viskell/ui/components/ValueBlock.java
```diff
@@ -22,16 +24,17 @@
  */
 public class ValueBlock extends Block {
     /** The value of this ValueBlock. */
-    private StringProperty value;
+    protected StringProperty value;
```
wandernauta: There are public `(get|set)value(property)` methods available. You could just use those from ValueBlock's subclasses.

Code/src/main/java/nl/utwente/viskell/ui/components/ValueBlock.java
```diff
@@ -52,6 +55,8 @@ protected ValueBlock(CustomUIPane pane, Type type, String value, String fxml) {
 
         outputSpace.setCenter(this.getOutputAnchor().get());
         outputSpace.toFront();
+        
+        this.valueType.setText("hoi");
```
wandernauta: Seems like debug code.

Code/src/main/java/nl/utwente/viskell/ui/components/ValueBlock.java
```diff
@@ -22,16 +24,17 @@
...
+    protected StringProperty value;
 
     /** The OutputAnchor of this ValueBlock. */
-    private OutputAnchor output;
+    protected OutputAnchor output;
 
     /** The space containing the output anchor. */
-    @FXML private BorderPane outputSpace;
+    @FXML protected BorderPane outputSpace;
+    @FXML protected Label valueType;
```
wandernauta: Missing Javadoc for `valueType` makes it seem it relates to the Javadoc for `outputSpace`.

Code/src/main/java/nl/utwente/viskell/ui/components/ValueBlock.java
```diff
@@ -88,6 +93,16 @@ public void updateExpr() {
...
+    
+    @Override
+    public void invalidateVisualState() {
+        try {
+			this.valueType.setText(this.expr.findType().prettyPrint());
+		} catch (HaskellTypeError e) {
+			this.valueType.setText("???");
+		}
+    	super.invalidateVisualState();
+    }
```
wandernauta: We use four-space indents, no tabs.

Code/src/main/resources/ui/style.css
```diff
@@ -65,14 +65,21 @@ CustomUIPane {
     -fx-text-alignment: center;
 }
 
-.display.block .argumentType {
-    -fx-padding: 0 5 0 5;
+.display.block .argumentType, .value.block .argumentType {
```
wandernauta: I'd say this is not an argument type (black-on-white) but a result type (white-on-black).

Flumble:
The colour of the valueblock types is reverted to black-on-white again, because the inverse is horrible, despite being more correct.

I believe subclasses (_SliderBlock_) should have more direct access to their properties (_ValueBlock::StringProperty_) than other classes _by default_. Right now the public getter and setter are simply returning the property as-is, but restrictions may be added in the future that shouldn't affect subclasses.

wandernauta:
If the inverse is horrible, why not change it elsewhere, as well? You've now introduced an inconsistency which would have to be explained ('white is in, black is out, except on such-and-such blocks').

As for the `protected` thing, I'd of course argue the contrary - subclasses should have as little access to *anything* as possible. In this case, SliderBlock should not depend on the fact that ValueBlock happens to have a property called 'value'. It should instead call ValueBlock's public API, which it seems it already does; ValueBlock should then be free to implement `getValue` and `setValue` in any way it sees fit. [ref](http://stackoverflow.com/a/2279692/182402)

comonoidial:
As for the background of types let's just make them all white for now, and figure out better visuals at a later stage.
I find it unproductive to discuss something minor as private vs protected in package being rewritten.

Issue #219, opened 2015-10-30
Lambda result/argument types (x_0) do not get turned into real types


wandernauta:
I'd expect every type outside the lambda block to just be `Float`.

![Screenshot](https://cloud.githubusercontent.com/assets/126832/10848734/9cd37ca8-7f1d-11e5-92d7-ffbded9da0b0.png)

comonoidial:
Solved by #221, but in the long term it might be better to have both an external and internal block(s) for each lambda and case like construct.

Issue #170, opened 2015-10-03
Resolve type checker bugs


Flumble:
The type checker right now is not bug-free. The main bug is that it crashes on recursion.

Some visual bugs:
- some functions like scanl *visually* have a signature (a -> a -> a) -> a -> [a] -> [a] when there's no wire connected (to fill in a type variable) and they are uncurried, though their actual signature is (b->a->b)->b->[a]->[b]
- some functions show a number as type variable instead of a letter (this number also increases with every change, so look into that)
- when currying/uncurrying a function for which the output is already attached to another node, this node may become invalid (shown with a red line and a warning sign), but it will not become *visually* valid again after moving the knot back.
- connecting an output with a second wire to a tree only *visually* updates 1 occurence of that type. For example, connect a True and an (==) to an (||) as inputs, then connect the True to the (==) as input –the type will read Bool only at the argument where the True is connected.

comonoidial:
While recursion is not supported now,  crashing should never happen.

Bug 1: is caused by the type variable implementation, and is fixed in my improved_types branch.
Bug 2: the number thing is solved by pull request #169, however the fast increasing typevar IDs shows that the UI is invoking the typechecker way too often.
Bug 3: looks like a true UI bug.
Bug 4: seems to be solved in the improved_types branch, but as soon as wires are disconnected it goes visually wrong again because not everything updates right.

Reproducing any of these bugs without using the UI in the unit tests would be appreciated.

comonoidial:
Any bugs left with the now fully rewritten typechecker, except for error connections as mentioned in #183?
Did you find anything else with wrong types, @Flumble ?

Flumble:
All the bugs are fixed except for a case resulting from scenario 4:
> connect a True and an (==) to an (||) as inputs, then connect the True to the (==) as input 

Viskell currently crashes (probably endless recursion, judging from the stack overflow) in an attempt to check this program in which a connection is made higher up in an expression tree (after which it is no longer a tree but a DAG). Note that this doesn't hold in general: for example, (type) manipulating a list and then zipping it with the original list works fine.

comonoidial:
Fixed in #221.

PR #221, opened 2015-10-30
Fix cornercases in the new updating strategy

comonoidial:
This should fix both #219 and #170.

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
@@ -3,6 +3,7 @@ package nl.utwente.viskell.ui.components;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
+import javafx.application.Platform;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.value.ObservableValue;
@@ -49,7 +50,7 @@ public abstract class Block extends StackPane implements Bundleable, ComponentLo
     protected BooleanProperty staleVisuals;
     
     /** Marker for the expression freshness. */
-    private boolean exprIsDirty;
+    protected boolean exprIsDirty;
 
     /**
      * @param pane The pane this block belongs to.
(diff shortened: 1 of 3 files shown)


Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -160,37 +161,47 @@ public void handleConnectionChanges() {
...
-                    index++;
-                }
-                // Get the Block in which the type error occurred and
-                // set the error state for the mismatched input to true.
-                getPane().getExprToFunction(errorExpr).getInput(index).setErrorState(true);
-                // Indicate that an error occurred in the latest analyze attempt.
-                getPane().setErrorOccurred(true);
-            }
+            // Needs to be delayed, because recomputation clears exprIsDirty also used to avoid infinite recursion.
+            Platform.runLater(this::recomputeExpression);
```
wandernauta: I don't really like this. `Platform.runLater` runs the given Runnable 'at some unspecified time in the future', according to the docs. Can we be certain that that unspecified time is 'late' enough to avoid the recursion?

comonoidial: All connected blocks get marked dirty immediately in handleConnectionChanges, and as long as runLater is the same gui thread it will be fine.
This was the least ugly solution i could find...

wandernauta:
This PR seems to introduce a regression.

*Steps to reproduce*: Create a lambda block for `fst (id x)`, then connect the result of `id` to the bottom anchor directly, which detaches the `fst`.
*Expected result*: `id` block shows `x_0` as input and output type (as implementation of lambda is now just `id`)
*Actual result*: `id` still shows `(a, b)` as input and output types.

Detaching the bottom anchor and re-attaching it shows `x_0` on the id block, as expected.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
comonoidial: Thinking about this more, when in the near future each block produces a let expr and the typechecker is per connection basis, then this workaround should not be needed anymore.

comonoidial:
I see, those pesky definition blocks break all kind of assumptions by having 3 sets of anchors.
Probably also for #222, these should be split into an inner and out block.

wandernauta:
Wouldn't splitting them add even more sets of anchors?

comonoidial:
The problem is that a definition block is both a block with 'normal' anchors as a block container with special internal anchors. So the concept of the block belonging to an anchor has multiple interpretations.
Same problem will show up for future case expression like constructs.

comonoidial:
This time a fix without new bugs I hope...

wandernauta:
Seems to be fixed. Thanks! :shipit:

PR #224, opened 2015-10-31
Rewrite of bezier curve point calculation

comonoidial:
Uses cubic root scaling factor instead of approximation, and short lines are more flexible now.


comonoidial:
I'm also considering letting wires with a large horizontal distance leave an anchor at an angle.

wandernauta:
Looks great! Thanks! :+1:

Issue #162, opened 2015-10-01
Change code generation strategy to a graph of let expressions


comonoidial:
Currenty haskell expressions for a program are generated in the from of a tree of nested expressions.
This has two big downsides the loss of sharing, and the expression tree can be exponentially large.
The solution is to generate expression globally from the block graph where each block becomes a let expression and each wire is associated with its own variable.
Also this will make implementing recursion and probably other constructions easier.

comonoidial:
Resolved by PR #253

PR #225, opened 2015-11-01
Add getChildren methods to Lambda and Annotated

wandernauta:
This makes both expressions introspectable in the inspector window.


comonoidial:
After #162 getChildren should be dumped, but it's practical for debugging now.

PR #228, opened 2015-11-01
Switch to connection based typechecking

comonoidial:
Work in progress, only the error detection and recovery uses it now, visualization still uses the expression tree types.
It's a big change thus early testing and comments welcome.
Also depends on #227.


comonoidial:
Now also the visualization uses the new connection based types, thus time for a bug hunt.

comonoidial:
I think it is all working now, please try and test.

Flumble:
Changes in connections within a lambda do not correctly propagate *visually* to the outside anymore.
In the example below, when attaching the wire in the top-right (highlighted in red), the ($) below will claim that the resulting type will be b instead of Num a:
($(\x -> ($) (+x) x))
Any further changes to the ($) will correct the type again. I don't see why this happens though, since any visual invalidation on the outside is (not?) set *after* the inside of the lambda is computed.

comonoidial:
@Flumble that bug is fixed now, and hopefull is the last.

PR #232, opened 2015-11-02
Tweak sizes of things for practicality or uniformity

comonoidial:
See the diff for trivial changes for convenience.


Code/src/main/java/nl/utwente/viskell/ui/ButtonOverlay.java
```diff
@@ -38,6 +38,7 @@ private FlowPane makeMenuBar(CustomUIPane pane) {
 
         FlowPane toolBar = new FlowPane(10, 0, menu);
         toolBar.setMaxHeight(40);
+        toolBar.setMaxWidth(40); // workaround to make it not confiscate the whole top of the screen
```
Flumble: This, and the maxHeight too, can be set to 0 for convenience, because the toolbar will always resize to fit its children (even if it gets larger than the window size).
With one icon it's already larger than the maximum of 40x40 pixels, so I'll merge it anyway.

PR #238, opened 2015-11-05
Add True and False to catalog (fixes #237)

### LINKED ISSUE #237: Add boolean constructors to catalog
The `True` and `False` data constructors would be useful to have in the catalog.

wandernauta:


wandernauta:
Replaced by #246.

PR #239, opened 2015-11-05
Remove Checkstyle from Maven config

wandernauta:
Nobody really pays any attention to Checkstyle's (many) warnings, which means it doesn't really add that much to keep around at this point.


comonoidial:
That clears up the build log nicely.

PR #246, opened 2015-11-06
Improvements and additions for catalog.xml

### LINKED ISSUE #237: Add boolean constructors to catalog

wandernauta:
Cleans up the XML a bit, and adds some functions and constructors for working
with tuples, booleans, Maybe and Either.

Replaces #238, fixes #237.

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ghcj/GhciEvaluator.java
+++ b/Code/src/main/java/nl/utwente/viskell/ghcj/GhciEvaluator.java
@@ -62,6 +62,11 @@ public class GhciEvaluator implements Closeable {
            useful sentinel. */
         this.eval(":set prompt " + SENTINEL);
 
+        /* Load some useful libraries. */
+        this.eval("import Data.List");
+        this.eval("import Data.Maybe");
+        this.eval("import Data.Either");
+
         /* Make it so that GHCi resets bindings after every command. This makes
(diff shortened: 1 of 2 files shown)


Code/src/main/java/nl/utwente/viskell/ghcj/GhciEvaluator.java
```diff
@@ -62,6 +62,12 @@ print a zero byte whenever it expects the user (that's us) to enter
            useful sentinel. */
         this.eval(":set prompt " + SENTINEL);
 
+        /* Load some useful libraries. */
+        this.eval("import Prelude");
+        this.eval("import Data.List");
+        this.eval("import Data.Maybe");
+        this.eval("import Data.Either");
```
Flumble: This can also be replaced by a single _":m Prelude Data.List Data.Maybe Data.Either"_.
Or even better, no literals, but just concatenating a(n observable) list of required modules.

wandernauta: Making the list observable sounds a bit scary for something so deep inside the backend, but you're right that this should be a list inside GhciEvaluator. Will fix.

wandernauta: Done.

PR #249, opened 2015-11-06
Make backend selectable at runtime

wandernauta:
Refactors the GhciEvaluator class into a tree of classes (Evaluator, PipeEvaluator, GhciEvaluator/ClashEvaluator). PipeEvaluator has most of the actual implementation; its superclass exists in case we find a cleaner way to talk to GHCi.

Also adds factory code to GhciSession that allows it to pick a backend at initialization time.

Adds a "Preferences" dialog to enable the user to pick their preference. The selection is stored as a `java.util.Prefs` value, and changing it restarts the GhciSession so that the preferred backend can be used immediately.

This PR is missing Javadoc, error handling, and tests, among other things, but I'd like to hear your thoughts anyway.


Code/src/main/java/nl/utwente/viskell/ghcj/ClashEvaluator.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @Override
+    protected List<String> getCommand() {
+        return ImmutableList.of("clash", "--interactive", "-ignore-dot-ghci");
+    }
+
+    @Override
+    protected List<String> getModules() {
+        return ImmutableList.of("+", "Data.List", "Data.Maybe", "Data.Either");
```
Flumble: "+" isn't exactly a module. :stuck_out_tongue:  I think it's better to put CLaSH.Prelude there*, so this is a list we can use to reload all the necessary modules (statelessly instead of relying on the ghci session) in the future when we get to adding and deleting module nodes.

*or enable a switch in CλaSH that implicitly loads its Prelude like GHCi does.
*or just move the "+" to the ":module " command.

wandernauta: Agreed. I think the first option makes the most sense; fixed.

comonoidial:
Clash is basicly a special version of ghc(i) with other library and compiler options, and with an extra vhdl generation command, so it should extend the ghci evaluator.
And I don't like adding abstract base classes/interface before there is a clear need with multiple instances.

wandernauta:
Changed the hierarchy up a bit: PipeEvaluator is gone, ClashEvaluator now extends GhciEvaluator which extends Evaluator. Javadoc, error handling, and tests are still pending.

Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
```diff
@@ -66,10 +67,8 @@ public CustomUIPane(HaskellCatalog catalog) {
         this.addEventHandler(KeyEvent.KEY_PRESSED, this::handleKey);
 
         /* Run the Inspector window. */
-        Platform.runLater(() -> {
-            inspector = new InspectorWindow(this);
-            inspector.blockProperty().bind(this.selectedBlock);
-        });
+        Platform.runLater(() -> inspector = new InspectorWindow(this));
+        Platform.runLater(() -> preferences = new PreferencesWindow(this));
```
comonoidial: Why use runLater if window creation can be done in the ShowPreferences() instead?

comonoidial:
Looks ok so far, however I don't have clash to test it.

Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
wandernauta: I'm not sure. Technically, I don't believe it's guaranteed that `showPreferences` is called on the JavaFX Application Thread, but in our case it might be. I'll check this out.

PR #250, opened 2015-11-07
The box and wire model with ports and components (WIP)

comonoidial:
This is my initial sketch of how the model could look like after ui model split.
Posted as PR for feedback, still lacks most methods and comments.


Code/src/main/java/nl/utwente/viskell/model/Component.java
```diff
@@ -0,0 +1,18 @@
+package nl.utwente.viskell.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Component extends Grouping {
```
wandernauta: Is there a more descriptive name than 'Component'?

Code/src/main/java/nl/utwente/viskell/model/Box.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.List;
+
+public interface Box {
+    
+    public Grouping getParent();
+
+    public List<InputPort> getInputs();
+
+    public List<OutputPort> getOutputs();
```
wandernauta: Is the caller of these methods allowed to modify the lists that are returned?

Code/src/main/java/nl/utwente/viskell/model/Component.java
comonoidial: Possibly, but I reused some terms from the hardware domain (ports, wires, component), and want to avoid name clashes with javafx.

Code/src/main/java/nl/utwente/viskell/model/ConstantBox.java
```diff
@@ -0,0 +1,43 @@
...
+    
+    private Grouping parent;
+    
+    private Expression expr;
+    
+    private Type type;
+    
+    private OutputPort ouput;
+
+    public ConstantBox(Component parent, String name, Expression expr, Type type) {
```
wandernauta: Would changing the value (expression) entail building and wiring up a new ConstantBox, or would `expr` be mutable? Having it mutable would perhaps help SliderBlock, but I'm not sure of the tradeoffs.

Code/src/main/java/nl/utwente/viskell/model/EvalBox.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Override
+    public List<InputPort> getInputs() {
+        return inputs;
+    }
+
+    @Override
+    public List<OutputPort> getOutputs() {
+        return new ArrayList<>();
```
wandernauta: If the result of `getOutputs()` is supposed to be immutable (see above), a sprinkling of ImmutableList's `of` and `copyOf`might be useful here.

Code/src/main/java/nl/utwente/viskell/model/FunctionBox.java
```diff
@@ -0,0 +1,33 @@
+package nl.utwente.viskell.model;
+
+import java.util.List;
+
+public class FunctionBox implements Box {
```
wandernauta: This is perhaps missing a reference to which function this FunctionBox is for.

Code/src/main/java/nl/utwente/viskell/model/Grouping.java
```diff
@@ -0,0 +1,14 @@
+package nl.utwente.viskell.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Grouping {
```
wandernauta: Can a Grouping have its own parent?

Code/src/main/java/nl/utwente/viskell/model/Wire.java
```diff
@@ -0,0 +1,14 @@
+package nl.utwente.viskell.model;
+
+public class Wire {
```
wandernauta: This seems at odds with use of the term 'Wire' in #251 - this would be represented by connections, not wires, in the UI, right?

comonoidial: Right, DrawWire in pr251 is not a good name...

wandernauta:
This looks very reasonable :+1: I can easily imagine building a graph of these, and I think it neatly represents everything we've got at the moment and everything we're planning to add.

I'm not yet sure where type checking (or even turning stuff into expression trees) would fit in, if at all, but that's mostly blocking on me grokking the connection-based type checking.

comonoidial:
What should be modifiable and not are open questions, and it not clear in the current ui code either.
And what methods are exactly needed depends on results expression graph stuff and other refactorings.
A big question for me is whether the model needs to/should refer back in some way to UI things.

wandernauta:
Yep. The introduction of these classes would be a good moment to settle who is allowed to change what and when.

I'm afraid that either model code will have to reference UI elements or vice versa, even if you'd ideally have neither.

Flumble:
Commits in [modelsplit](https://github.com/wandernauta/viskell/commits/modelsplit) can also be found in [modelsplit-wn](https://github.com/wandernauta/viskell/commits/modelsplit-wn).

PR #253, opened 2015-11-08
Switched code generation to a graph of let expressions

comonoidial:
This is quite a rewrite and it also required changing the update strategy again.
PR #252 should be merged first.
Finally issue #162 is resolved.


comonoidial:
This PR is ready for review, but I will keep adding follow up refactorings as long as nobody starts a review.

PR #254, opened 2015-11-08
Refactored anchors, to avoid duplication/inconsistencies with binder types and anchor types

comonoidial:
Also depends on #253.


comonoidial:
Resolved in PR #253.

comonoidial:
Oops had wrong window open...

PR #258, opened 2015-11-11
Split in/outputAnchors graphics and add extra wire to unconnected ones

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/components/OutputAnchor.java
```diff
@@ -18,6 +20,15 @@
...
     
+    /** The visual representation of the OutputAnchor. */
+    @FXML private Shape visibleAnchor;
+    
+    /** The invisible part of the OutputAnchor (the touch zone). */
+    @FXML private Shape invisibleAnchor;
+
+    /** The thing sticking out of an unconnected OutputAnchor. */
+    @FXML private Shape openWire;
+
```
wandernauta: Duplication with `InputAnchor` (why move it in the first place?)

wandernauta:
Can you explain your reasoning here? At first glance, this PR seems to just introduce duplication without bringing any real benefit.

The extra wire looks a bit odd when making new connections:

screenshot

comonoidial:
Both anchor types will become increasingly different because of extra wire, later integration with type labels that hides in certain situations.
The visuals can use a little tweak indeed.

wandernauta:
Would it make sense to split them when they become different, instead of in advance? The open wire doesn't really count as a difference; the difference between input and output is just two minuses, something which could just as easily be solved some other way, for example with FX CSS.

comonoidial:
It could be done later, just preparing for plannen future changes, reducing the differences with the model, and working towards making connectionAnchor merely an interface. Also I don't think of css as a solution...

wandernauta:
Okay.

Do you want to fix the openWire disappearing too late in this PR, or merge as is?

comonoidial:
Merge it as is, I rather not spent much time on visual cornercases this month, while much basic functionality is still missing and all will still change a lot.

wandernauta:
Done, thanks!

PR #260, opened 2015-11-12
Make Block more abstract and related cleanups

comonoidial:
Also simplifies visual state updates a bit

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/CircleMenu.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/CircleMenu.java
@@ -1,9 +1,11 @@
 package nl.utwente.viskell.ui;
 
+import java.util.ArrayList;
+
 import javafx.scene.control.MenuItem;
 import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
-import javafx.scene.layout.StackPane;
+import javafx.scene.input.MouseEvent;
 import javafx.util.Duration;
 import jfxtras.scene.layout.CircularPane;
 import jfxtras.scene.menu.CirclePopupMenu;
@@ -22,11 +24,23 @@ import nl.utwente.viskell.ui.components.Block;
  */
 public class CircleMenu extends CirclePopupMenu {
 
+    /** List of all block on which a CircleMenu is active. */
+    private static ArrayList<Block> activeBlocks = new ArrayList<>();
+    
     /** The context of the menu. */
     private Block block;
 
+    /** Show the Circle menu for a block if not already active. */
+    public static void showFor(Block block, MouseEvent t) {
+        if (! activeBlocks.contains(block)) {
+            activeBlocks.add(block);
+            CircleMenu menu = new CircleMenu(block);
+            menu.show(t);
+        }
+    }
+    
     public CircleMenu(Block block) {
-        super((StackPane) block, null);
+        super(block, null);
         this.block = block;
 
         // Define menu items
@@ -65,6 +79,12 @@ public class CircleMenu extends CirclePopupMenu {
         return image;
     }
 
+    @Override
+    public void hide() {
+        CircleMenu.activeBlocks.remove(this.block);
+        super.hide();
+    }
+    
     /** Copy the {@link Block} in this context. */
     public void copy() {
         // TODO implement clipBoard in main app.
--- a/Code/src/main/java/nl/utwente/viskell/ui/CircleMenu.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/CircleMenu.java
@@ -1,7 +1,5 @@
 package nl.utwente.viskell.ui;
 
-import java.util.ArrayList;
-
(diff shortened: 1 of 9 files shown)


Code/src/main/java/nl/utwente/viskell/ui/CircleMenu.java
```diff
@@ -22,11 +24,23 @@
  */
 public class CircleMenu extends CirclePopupMenu {
 
+    /** List of all block on which a CircleMenu is active. */
+    private static ArrayList<Block> activeBlocks = new ArrayList<>();
```
wandernauta: While I can't come up with any ways it could actually go wrong, I have a bit of a dislike for global, mutable state like this. Is there a better way to store this? Maybe a block could keep track of whether it has a circlemenu on it or not?

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -84,11 +68,7 @@ public Block(CustomUIPane pane) {
     private void handleMouseEvent(MouseEvent t) {
         parentPane.setSelectedBlock(this);
         if (t.getButton() == MouseButton.SECONDARY) {
-            if (this.circleMenu == null) {
-                this.circleMenu = new CircleMenu(this);
-            }
-            
-            this.circleMenu.show(t);
```
wandernauta: Ah, like this. Let the Block keep track of its (singular) menu.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -169,33 +145,25 @@ protected void propagateConnectionChanges(boolean finalPhase) {
...
-            this.updateExpr();
-        }
-
         // propagate changes down from the output anchor to connected inputs
         this.getOutputAnchor().ifPresent(a -> a.getOppositeAnchors().stream().forEach(x -> x.handleConnectionChanges(finalPhase)));
 
-        // If the change is not propagated any further down start recomputation.
-        if (finalPhase && this.isBottomMost()) {
-            // Now that the expressions and types are updated, initiate a visual refresh.
-            Platform.runLater(() -> this.staleVisuals.set(true));
```
wandernauta: This hunk and the one above seem to be more of an actual behavioural change than a cleanup - are you sure they do the same thing in all cases?

comonoidial: This is a simplification now that the order of invalidating visual state doesn't matter anymore.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -100,17 +80,13 @@ public final CustomUIPane getPane() {
...
-    }
+    public abstract List<InputAnchor> getAllInputs();
 
     /**
      * @return the optional output Anchor for this Block
      * TODO generalize to List<OutputAnchor> getOutputAnchors()
      */
-    public Optional<OutputAnchor> getOutputAnchor() {
-        return Optional.empty();
-    }
```
wandernauta: I'd say these are reasonable defaults - I'm curious about your reasons for dropping them.

comonoidial: Just disliking the arbitrarily overriding of methods, I think almost all should be either abstract or final. It should be obvious what a subclass should override.

Code/src/main/java/nl/utwente/viskell/ui/CircleMenu.java
comonoidial: This is fixing up a corner case in circlemenu behaviour, and I think that code does not belong/has anything to do with Block.

wandernauta: That's true, but "a block has-a menu" makes more sense to me than "the concept of circle menus has-a set of blocks". I'll have to think about this.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
wandernauta: Okay.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
wandernauta: Seems to work, neat! :+1: You're right, it really does simplify things.

Flumble:
I don't really see the benefit of limiting blocks to one CircleMenu (if the user wants to open multiple context menus, let them), so you can leave the list (or property) out of the code by me.

wandernauta:
If I understand it correctly, the new list is there to prevent multiple circle menus being opened on the same block.

comonoidial:
I'm fine with removing that list instead.

wandernauta:
Me too :+1:

wandernauta:
Merged, thanks!

PR #262, opened 2015-11-14
Split Defintion block into a inner and outer part 

comonoidial:
Also types now only propagate outwards from a lambda.
Depends on #260.


wandernauta:
This change seems to have shifted the anchors a few hundred px:

screenshot

comonoidial:
Yes, internal anchors are now centered because it looks ugly and unexpected on the left side.

wandernauta:
I'll assign this PR to @Flumble because it contains visual changes, as discussed.

wandernauta:
Fantastic! Merged. :sparkles:

PR #263, opened 2015-11-15
Make GhciSession a Service that returns Futures

wandernauta:
This enables GhciSession to execute Haskell snippets on a different
thread than the main thread.

GhciSession is an AbstractExecutionThreadService from Guava. It takes Haskell
source to execute off the work queue, which is read from a (singular)
execution thread, and filled by anything that wants something executed.

Push and pull methods on GhciSession return 'immediately', and never
throw HaskellException. If (when) something goes wrong, the future will
be set to unsuccesful, which means `get` will throw ExecutionException
or the unsuccesful callback will be called.

DisplayBlock uses callbacks to delay updating of its label until the
future is completed (i.e. evaluation should not block the UI).

RGBBlock and GraphBlock wait for their results for now (i.e. their
behaviour should be unchanged and they will block the UI).

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/components/DisplayBlock.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/components/DisplayBlock.java
@@ -101,11 +101,13 @@ public class DisplayBlock extends Block {
 
                 Futures.addCallback(result, new FutureCallback<String>() {
                     public void onSuccess(String s) {
+                        // Can't call setOutput directly - this may not be JavaFX app thread.
+                        // Instead, schedule setOutput to be done some time in the future.
                         Platform.runLater(() -> setOutput(s));
                     }
 
                     public void onFailure(Throwable throwable) {
-                        Platform.runLater(() -> setOutput("?!?!?!"));
+                        onSuccess("?!?!?!");
                     }
                 });
             }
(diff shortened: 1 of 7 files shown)


Code/src/main/java/nl/utwente/viskell/ui/components/DisplayBlock.java
```diff
@@ -90,14 +94,21 @@ public void invalidateVisualState() {
...
-                setOutput("?!?!?!");
-            } 
+            Optional<GhciSession> ghci = getPane().getGhciSession();
+
+            if (ghci.isPresent()) {
+                ListenableFuture<String> result = ghci.get().pull(inputAnchor.getFullExpr());
+
+                Futures.addCallback(result, new FutureCallback<String>() {
+                    public void onSuccess(String s) {
+                        Platform.runLater(() -> setOutput(s));
```
comonoidial: What is the purpose of a runLater on a call back?

wandernauta: The callback gets called on *some* thread (which doesn't have to be, and in this case probably isn't, the JavaFX Application Thread), while the JavaFX Application Thread is the only thread that is allowed to call setOutput. Platform.runLater, of course, schedules work on the JavaFX Application Thread.

A neater solution would be if we could somehow get an `Executor` that corresponds to the JavaFX Application Thread and pass that to `addCallback`, but I haven't been able to find anything like that.

comonoidial: I see, maybe add a litte comment here so the next person doesn't remove it.

wandernauta: Good idea! Done.

Issue #230, opened 2015-11-02
Make the function menu less annoying to use


comonoidial:
This needs no further comments, just trying to use it will show multiple shortcomings.

wandernauta:
Neverless, specifying the shortcomings could make for a clearer issue report, as it's a bit vague as is.

comonoidial:
Let's try to make a start:
* it wastes too much space
* function categories should not expand down in an already vertical menu
* the menu should not appear partially offscreen
* also should not expand into offscreen
* please no scrollbars
* expanded categories should not have empty lines
* it is visually inconsistent
* the current grouping in categories itself is of little use, unless you know what to find where
* alphabetic ordering is not good with many rarely used items and lots of operators
* uncommon blocks should not be on the first row

wandernauta:
A way to deal with much of the positioning and layout stuff could be to use a regular JavaFX ContextMenu, at least on non-touch devices, as it automatically positions itself in a sensible way so that it doesn't move offscreen. Another advantage is that it has a consistent size that doesn't depend on the zoom level, and that people are used to the way it works.

The grouping and ordering things could perhaps be approached, again at least on non-touch devices, by adding (prefix-, type- or fuzzy) keyboard-based search.

comonoidial:
And I forgot one important thing, that is being able to drag new block from the function menu to the location you want them. Having a click menu stack them all in one spot is very inconvenient.

comonoidial:
Oops did overlook #164, let's continue here.

Flumble:
From https://github.com/wandernauta/viskell/pull/304#issuecomment-163381402

  * The margins around the buttons could perhaps use some fiddling around with to make them look neater
  * I get the extra Close button from a balance standpoint, but I don't think it's good UX - if people see a close button at that position, they will get used to it being there. 'Suddenly' hiding it when the user switches backend would then not be ideal. I'd prefer to just leave the space blank (or purple).
  * ...aaand when you're playing around with the FunctionMenu css anyway - maybe the Close button could use a splash of color to help distinguish it from the others?

comonoidial:
I think basicly all that can be improved with this style of function menu has been done.
Further non trivial improvement will require a complete redesign.

Issue #229, opened 2015-11-02
Support functionless function blocks


comonoidial:
At the moment you have first select an exact function from the less practical function menu, and then connect it to where you want to use it, but often you would like to do that in other way around.
So starting with an functionless block with a certain number of arguments to connect, and then after connecting the inputs, use the avaible types to select a fitting function for this block (filtered to a convenient small set based on types).

wandernauta:
This could be the apply-anything block I drew in #194. You'd attach your arguments, then maybe touch the spot where the function name is supposed to go to get a list of matching functions, as in #223. Picking a function would turn the block into a regular function block.

comonoidial:
Having thought about this more, this is indeed as #194 with a little extra user interface.

Issue #223, opened 2015-10-31
Context menu for wires that are partially connected


comonoidial:
While creating connections it can be convenient to insert a new block instead on the open end of a wire.
At this moment we know already a type and whether it is an input or output, thus can give the user a much more specific selection of block to insert at the current position.
How to activate this menu by touch is an open question.

comonoidial:
This requires #206 to have a good way to catch events for showing this menu.

comonoidial:
Resolved in #347.

PR #265, opened 2015-11-15
Add getBy{Predicate,Prefix,Type} to HaskellCatalog

wandernauta:
These methods return a subset of the functions in the catalog, by
filtering on a user-supplied predicate, prefix, or type, respectively.

Intended as groundwork for #230, #229, #223.


wandernauta:
Investigating Travis failure.

wandernauta:
Rebased for your pleasure.

comonoidial:
Looks fine.

Issue #243, opened 2015-11-05
Combine anchors and type labels


comonoidial:
Most anchors already have an associated type label, however implemented in various ad hoc ways.
For consistency every anchor should have a type label, and the model behind the ui every anchor already has a type.
And visually the relation can be made stronger by moving the label to the edge of a block.
This will also replace the special inputargument class for function blocks.

comonoidial:
It looks like combining them causes all kinds small practical issues, thus it's not worth it.

PR #266, opened 2015-11-15
Simplify the fxml of Display/Value/Slider-Blocks

comonoidial:


wandernauta:
This PR (again) changes the appearance of some blocks - ValueBlock and DisplayBlock have some extra padding now, but SliderBlock doesn't. Is that intentional?

comonoidial:
Not sure where that padding comes from, and its not so important before #243.

comonoidial:
I will stop adding things to this PR, even though lots of ugliness is left.

Flumble:
The padding in the display and value blocks appears because the nodes inside are smaller than the block and they get tucked together in the middle.
I approve of the changes and have some additional ones:

wandernauta:
Which additional ones? You have my curiousity

comonoidial:
@Flumble this has some minor conflicts due to merging of #272.

Flumble:
The conflicts are resolved now.

Issue #157, opened 2015-09-29
Add support for keyboard hotkeys


Flumble:
Currently there is keyboard support to translate and scale the view, to inspect nodes and to remove nodes.
More keyboard hotkeys are desired, e.g. for selecting/traversing nodes and for adding new nodes.

PR #268, opened 2015-11-16
Drop the single global selected block property

comonoidial:


wandernauta:
This seems to conflict a bit with #157.

comonoidial:
A global selected block does not make sense in a multitouch environment.
Making keyboard support only a secondary backup option is a very deliberate design choice.
It forces us to think hard about how to design a practical interactive UX, instead of taking the easy old route of typing textual things.

wandernauta:
I guess touches on a design decision that hasn't been properly made yet. The way I see it, the choice is between...

* Viskell is *also* designed with touch interaction in mind, that is, it (has a mode in which it) can be used without mouse and keyboard.
* Viskell is *primarily* designed for touch, but those stuck in ancient traditions like yours truly can use mouse and keyboard and be productive.
* Viskell is *only* intended for people with Windows machines that have large touchscreens.

It might be a good idea to discuss this more extensively in the next meeting.

Flumble:
Since we forgot about this in the last meeting, I'll join the discussion here.

In my opinion touch interaction is our primary goal and interaction via other means (mice, which have multiple buttons and a hover, or keyboards, which have lots of buttons) is a nice addition whenever you feel like supporting it. But it shouldn't interfere with the touch interface.

For "those stuck in ancient traditions" there are loads of IDEs available already. (The only thing missing to bring these worlds together is a means to export and import Haskell code)

Re this PR: The block selection in its current form convolutes touch and keyboard interaction.

PR #272, opened 2015-11-17
Added preferences option for the UI style

Flumble:
Currently simply a fixed set of paths to the CSS files. A theme class or
map can be made later on. Likewise a set of all the existing stages so
you don't have to hard-code replacing the style of each stage.
(That's what the commit says already.)


Flumble:
Can you test if it reloads properly by switching the style twice?
On my system JavaFX loads the old CSS file until I restart the application completely.

comonoidial:
That works fine for me, maybe you should update your system...

PR #274, opened 2015-11-18
Fixed dragging nodes

Flumble:
and preventing the function menu from being dragged along


wandernauta:
This looks reasonable :+1: 

Does it also work without the `event.consume`?

Flumble:
If it doesn't get consumed, the synthesized event will bubble to the background (or any parent for that matter), since the dragContext is already set, and initiate a drag action there.

comonoidial:
Seen it work on the big touch screen :+1:

PR #276, opened 2015-11-19
Add show constraint to the input of display blocks

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/components/DisplayBlock.java
```diff
@@ -101,7 +100,7 @@ public Expression getLocalExpr() {
     
     @Override
     public void refreshAnchorTypes() {
-        this.inputAnchor.setFreshRequiredType(TypeScope.unique("any"), new TypeScope());        
+        this.inputAnchor.setFreshRequiredType(this.getPane().getEnvInstance().buildType("Show a => a"), new TypeScope());
```
wandernauta: Would it make sense to have DisplayBlock keep an instance of this type and keep making fresh copies of it? This invokes the ANTLR parser every time.

comonoidial:
Probably doesn't make a big difference, but changed anyway.

wandernauta:
Great, thanks!

PR #278, opened 2015-11-19
Fix format of slider block floats to be always valid.

comonoidial:


wandernauta:
Travis failure seems to be related - missing import, perhaps?

Flumble:
As much as I dislike forcing a locale, it's consistent this way with the required notation in value blocks and displays.

PR #279, opened 2015-11-21
Refactor and integrate all reusable parts of tactilefx into the ui package

comonoidial:


wandernauta:
Thanks! I'll take a look.

Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
```diff
@@ -61,6 +62,12 @@ public CustomUIPane(HaskellCatalog catalog) {
         this.addEventHandler(KeyEvent.KEY_PRESSED, this::handleKey);
     }
 
+    /** @return modifiable list of children. */
+    @Override 
+    public ObservableList<Node> getChildren() {
+        return super.getChildren();
+    }
+
```
wandernauta: Why do we need to override this?

Code/src/main/java/nl/utwente/viskell/ui/DragContext.java
```diff
@@ -0,0 +1,181 @@
...
+        /**
+         * Represents processing events at the handler stage
+         */
+        HANDLER, 
+        
+        /**
+         * Represents processing events at the filter stage.
+         */
+        FILTER
+    }
```
wandernauta: Is this distinction still important when we're the only users of the code?

Code/src/main/java/nl/utwente/viskell/ui/DragContext.java
```diff
@@ -0,0 +1,181 @@
...
+    public static final int MOUSE_ID = -2;
+    
+    final Node node;         // Node that is being dragged
+    /* Whether this node will go to the foreground when the user starts a drag gesture with it. */
+    boolean goToForegroundOnContact; //
+    double localOffsetX, localOffsetY;  // The x,y position of the Event in the Node
+    int touchId;            // The id of the finger/cursor that is currently dragging the Node
+    
+    private final EventHandler<TouchEvent> touchHandler;
+    private final EventHandler<MouseEvent> mouseHandler;
```
wandernauta: These fields could perhaps use Javadoc comments.

Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
comonoidial: Because it is protected by default, and we directly manipulate UIPane parts (which is a bit ugly).

Code/src/main/java/nl/utwente/viskell/ui/TouchDisplay.java
```diff
@@ -32,16 +35,21 @@ public TouchDisplay(double x, double y, double radius, int touchId) {
...
+        
+        line = new Line(x, y, bounds.getMinX(), bounds.getMinY());
+        getChildren().add(line);
 
         setBackground(new Background(new BackgroundFill(Color.TRANSPARENT, null, null)));
     }
 
-    @Override
-    public void relocate(double x, double y) {
+    public void relocate(double x, double y, Bounds bounds) {
```
wandernauta: This no longer overrides `relocate`. Would calling the non-overridden `relocate` do subtly the wrong thing? If so, I would be in favor of renaming this method.

wandernauta:
I haven't tested this yet as I'm on the move, but I like this refactor a lot. Fantastic!

Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
wandernauta: Ah, gotcha. It's public in `Pane` but protected in `Parent` (and therefore `Region`).

comonoidial:
This will require testing on the big touch screen before merging @Flumble

comonoidial:
Tested on the big screen and ready for pull.

PR #281, opened 2015-11-25
Shift anchors to the outside of blocks

comonoidial:


wandernauta:
Assigning to @Flumble as this has visual changes.

wandernauta:
Code-wise there's not much to say anything about, but I've got a few questions on the visual changes:

- The black anchors on a dark gray background are a bit hard to see. Have you tried light gray anchors?
- The transforms seem slightly off for (at least) SliderBlock: its output anchor goes one pixel into the block's body. Is this intentional/does this mean anything?

comonoidial:
I'd rather have dark anchor/wire on a light background (see #216), so that things are readable also in a presentation on a beamer. Maybe you could make an alternative css color style, more suitable for night time hacking.
Some small offset problems exist and it seems to have something to do with how border widths are effecting things.

PR #282, opened 2015-11-26
Alternative function application blocks with vertical currying

comonoidial:
For now hidden under the right mouse button, as it needs some more testing and visual tweaks.


Code/src/main/java/nl/utwente/viskell/ui/components/FunApplyBlock.java
```diff
@@ -0,0 +1,224 @@
...
+                @Override
+                public void relocate(double x, double y) {
+                    super.relocate(x, y);
+                    FunInputAnchor.this.dragShift(y);
+                }
+            };
+            this.inputType.setMinWidth(USE_PREF_SIZE);
+            this.inputType.getStyleClass().add("inputType");
+            this.anchor = new InputAnchor(FunApplyBlock.this);
+            this.anchor.layoutXProperty().bind(this.inputType.widthProperty().divide(2));
```
wandernauta: I'll allow using a bind here - it's simple enough and there's only one of them - but officially prefer getting it right with layouts.

Code/src/main/java/nl/utwente/viskell/ui/components/FunApplyBlock.java
```diff
@@ -0,0 +1,224 @@
...
+            FunType ft = (FunType) t;
+            FunInputAnchor ia = new FunInputAnchor();
+            this.inputs.add(ia);
+            t = ft.getResult();
+        }
+
+        Pane inputSpace = new HBox(10, this.inputs.toArray(new Node[this.inputs.size()]));
+        this.bodySpace.getChildren().addAll(inputSpace, outputSpace);
+        outputSpace.layoutYProperty().bind(inputSpace.heightProperty());
+        outputSpace.layoutXProperty().bind(inputSpace.widthProperty().divide(2).subtract(resTypeLabel.widthProperty().divide(2)));
```
wandernauta: What is this even trying to do?

wandernauta:
The code looks reasonable. As you know or no doubt have guessed, I'm not in favour of the visual changes; I'm not sure if I've written down my motivations anywhere yet, so here goes.

- I personally don't know how to parse the visual syntax:  
screenshot  
This looks like a function that takes one number (so far so good) and returns one number (eh?). The other input that was curried obviously isn't an input (as it has no anchor attached to it) and because it's on the bottom, but it's also not an output (it's not white-on-black).
- I'm worried that, if I don't know how to parse it without instruction, people who are new to Viskell will have an even harder time. However, I do not have any numbers to back this up.
- I also don't really get the whole idea of currying, say, the second argument of a four-argument function. I can imagine it being useful in some cases but I've never had to do it in my (admittedly limited) Haskell programming so far. I have used `flip` once or twice, but that's already in the catalog.
- This turns "curry the last three arguments" into three actions (swipes) instead of one.
- I'm (usually) stubborn and (occasionally) conservative and LwB-style currying has been in there since June. I'm just starting to like it.

Code/src/main/java/nl/utwente/viskell/ui/components/FunApplyBlock.java
comonoidial: It is simply centering one thing to center of another thing.
I searched for hours for better solution with standard layout blocks, but didn't find any.
The problem is with how event handling is coupled with node/parent hierachy, and that doesn't work well with moving parts.
And workarounds with setMouseTransparent don't work either, because that also affect all (even indirect) children nodes.

wandernauta: Okay, I'll take your word on this.

comonoidial:
* This feature is intentionally hidden under the right mouse button, so that we can compare and gather feedback from other people.
* It is visually not finished yet, will look into improving that this weekend.
* I think this vertical currying fits better with in the 2D (not linear expression) based structure of viskell.
* Making all arguments 'equals' in currying resolves the tricky library design question of argument order.
* That it costs more actions in rare cases is a minor issue compared to unstable resizing swiping knots.
* If getting used to it is an argument, may I take in account the time thinking about the design before even OWP started :wink:

wandernauta:
I don't immediately see us reaching a consensus on this and I wouldn't want to stuff this behind a preference. I'll re-assign to @Flumble.

Flumble:
I'm in favour with everything except for the output label. Like Wander pointed out in https://github.com/wandernauta/viskell/pull/282#issuecomment-160132044 , it is not very clear what the output type of a curried function is, so I'd rather see the whole output type again or have the output type to the right of all the curried arguments (perhaps including an arrow between them).

comonoidial:
I would like to see this merged as is, where the vertical currying is only a secondary alternative.
Clearly this needs more work before it is ready as main option, but having a seperate branch is quite inconvenient when many planned changes affect both kind of function blocks.

PR #283, opened 2015-11-27
Make the custom uipane layered

### LINKED ISSUE #269: Split graphics into multiple layers
Currently all graphical elements are one big list of things, keeping the stacking order look good and usable becomes increasingly hard.
A solution is to divide graphical elements over a few layers:
The bottom layer will be block 'containers' such as lambdas, cases and current toplevel.
Next layer are the normal blocks as functions, values and displays.
Then a layer with all the wires connecting things without being accidently obscured.
And the top layer will be menu like things and blocks be dragged around.
One complication is that currently tactilefx doesn't deal well with hierachy.

comonoidial:
Resolves #269

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
@@ -12,6 +12,7 @@ import nl.utwente.viskell.ghcj.GhciSession;
 import nl.utwente.viskell.haskell.env.Environment;
 import nl.utwente.viskell.haskell.env.HaskellCatalog;
 import nl.utwente.viskell.ui.components.Block;
+import nl.utwente.viskell.ui.components.DefinitionBlock;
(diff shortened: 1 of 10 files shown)


wandernauta:
This seems to work great, thanks! Two minor things:

- Maybe you could consider ask a Block class what layer it thinks it should be on instead of hardcoding the is-it-a-DefinitionBlock-subclass criterion.
- I don't really like the hierarchical layer-building, esp. the skipping of one index in `streamChildren`. Wouldn't it be easier to have CustomUIPane be a StackedPane or something similar, and the layers just be on top of each other instead of containing one another?

comonoidial:
On the first point that would be nicer indeed.
And this unusual layering construction is unfortunare result of how javafx event handling works.
I could not find another way of building this where events still arrive reliably at the bottom layer nodes.

wandernauta:
That's exactly what I had in mind, great! And of course, event handling. That makes sense.

Will merge, thanks again!

PR #284, opened 2015-11-28
Make lambda blocks resizable

comonoidial:


wandernauta:
The code looks reasonable. I'm not sure I like the color, and I'm not sure I like that it's not controllable through CSS, but it seems to work well.

An alternative solution would be to resize the lambdablock to the bounding box of its children. This is when Blender does for group boxes. A partial implementation of this is in the `autoresize-lambda-block` branch. Would you prefer this approach, that approach, or both approaches?

comonoidial:
I didn't spent much time on how it looks, that something of lower priority for later.

The bounding box approach is good for finished program parts and setting minimum sizes.
When constructing a program you might want to reserve space, and resize the lambda to bounds where it is easy to drag blocks into particular places.
So ideally it is a combination of both.

wandernauta:
Agreed. Will merge, thanks again!

PR #287, opened 2015-12-01
Add a special catalog for CLaSH

wandernauta:
CLaSH has its own functions, types and (soon) blocks. It therefore needs its own catalog. This PR adds a first (incomplete) version of that catalog, and a mechanism for backends (evaluators) to let Viskell know which catalog they need.


wandernauta:
I can rebase+squash this PR if preferred.

PR #288, opened 2015-12-02
Drop error image and use error coloring on the anchor instead

comonoidial:


wandernauta:
Looks good - the System.err call seems like debug code, other than that feel free to merge

PR #289, opened 2015-12-03
Debug improvements

Flumble:
Just a quickie.
Having the inspector window pop up in the middle of the screen is annoying.


wandernauta:
Merged, thanks!

PR #290, opened 2015-12-03
Tweaked looks and placement of wires and anchors

comonoidial:


wandernauta:
I like the smoother lines, they really look quite natural.

However, I'm not the biggest fan of making connections black on a dark-grey background. Will assign to @Flumble per the agreement.

PR #293, opened 2015-12-04
Add first version of SimulateBlock

wandernauta:
> SimulateBlock is CLaSH-specific block that is a counterpart to the `simulate`
> function in Clash. Expects a function that turns a Signal of monotonically
> increasing numbers into a Signal of representable result values.
> 
> While signals are infinite, SimulateBlock only shows the first N
> iterations/steps.

It's basically a poor man's (stepping) debugger.

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/components/SimulateBlock.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/components/SimulateBlock.java
@@ -42,6 +42,9 @@ public class SimulateBlock extends Block implements ComponentLoader {
 
     @FXML protected Button iterationLabel;
 
+    /** Constrained type variable for the input anchor */
+    private final Type funConstraint;
+
     public SimulateBlock(CustomUIPane pane) {
         super(pane);
         loadFXML("SimulateBlock");
@@ -52,6 +55,9 @@ public class SimulateBlock extends Block implements ComponentLoader {
         iteration = new SimpleIntegerProperty(0);
         iteration.addListener(e -> this.invalidateVisualState());
         iteration.addListener(e -> iterationLabel.setText(String.valueOf(iteration.get())));
+
+        String signature = "(Num a, Show b) => Signal a -> Signal b";
+        funConstraint = getPane().getEnvInstance().buildType(signature);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


Code/src/main/java/nl/utwente/viskell/ui/components/SimulateBlock.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public Expression getLocalExpr() {
+        return inputAnchor.getLocalExpr();
+    }
+
+    @Override
+    public void refreshAnchorTypes() {
+        String signature = "(Num a, Show b) => Signal a -> Signal b";
+        Type type = getPane().getEnvInstance().buildType(signature);
```
comonoidial: The type could be built once in the constructor

Code/src/main/resources/ui/layout.css
```diff
@@ -107,3 +107,10 @@ Connection {
     -fx-background-insets: 0, 0;
     -fx-background-radius: 0, 0;
 }
+
+.subtle {
+    -fx-background-color: rgba(0, 0, 0, 0);
+    -fx-font-size: 12px;
+    -fx-font-weight: bold;
+    -fx-text-fill: white;
+}
```
comonoidial: newline?

Code/src/main/resources/ui/SimulateBlock.fxml
```diff
@@ -0,0 +1,27 @@
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.layout.*?>
+<?import nl.utwente.viskell.ui.components.SimulateBlock?>
+<?import javafx.scene.control.Button?>
+<fx:root type="nl.utwente.viskell.ui.components.SimulateBlock" xmlns:fx="http://javafx.com/fxml/">
+    <BorderPane styleClass="display, block">
+        <center>
+            <VBox alignment="CENTER">
+                <!-- silly extra HBox to not break the argumentSpace css -->
```
comonoidial: Maybe someone knowing about css could fix this instead of making more copies of a workaround.

Code/src/main/resources/ui/SimulateBlock.fxml
```diff
@@ -0,0 +1,27 @@
...
+                <HBox styleClass="argumentSpace">
+                    <Label fx:id="inputType"/>
+                </HBox>
+                <VBox>
```
comonoidial: Why is this nested VBox needed?

comonoidial:
Code looks fine except for some trivial things.
However I do not understand the choice for an integerproperty with a listener, while recalculation could also be done in a setter for the iteration count.

wandernauta:
You're right, we don't need a property there.

Comments should be fixed now.

PR #295, opened 2015-12-05
Add information about datatypes to the catalog

comonoidial:


wandernauta:
The Java code looks perfectly fine at first glance.

However, is this the most efficient way to represent types? Would it make sense to add information about which types are in which classes to the `datatypes` node and get rid of the `classes/class/instance` nodes? It might get rid of some duplication in the parser, at least.

comonoidial:
Typeclasses are a more general concept than just interfaces on datatypes, and instances can be added later even in other modules.
So nothing to simplify here, without restricting future catalog extensions.

wandernauta:
And an `instance`'s `constrainedArgs` can't be deduced from a `dataType`'s `typeArity`, for example?

comonoidial:
No, that may be any number between 0 and the type arity.

wandernauta:
Good point. Will merge, thanks!

PR #296, opened 2015-12-06
Add secondary action to dragcontext and use it for circlemenu

comonoidial:
This should make a fast 'rightclick' possible for touch, but that needs testing on the big screen.


comonoidial:
Has been tested on the big screen.

wandernauta:
Merged, thanks!

PR #297, opened 2015-12-06
Use value blocks for things without arguments

comonoidial:


wandernauta:
This makes sense; will merge, thanks!

PR #298, opened 2015-12-07
Case block and containers

Flumble:
Changes:
  - Preliminary case block added to the menu consisting of a surrounding block with an output anchor and 2 alternatives with a result anchor (adding and removing alternatives is supported by the block)
  - Alternatives in a case block generate an expression in the form of a pattern, a pattern guard and a body (though in its current form only the guard is populated dynamically and the body is populated by the result anchor)
  - Blocks can attach themselves to a container and expression building is limited accordingly
  - Expression building now returns both the expression and a set of required blocks for that expression


wandernauta:
Thanks, I'll take a look.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -151,46 +162,73 @@ public void handleConnectionChanges(boolean finalPhase) {
     /**
      * @return The local expression this Block represents.
      */
-    public abstract Expression getLocalExpr();
+    public abstract Pair<Expression,Set<Block>> getLocalExpr();
```
wandernauta: Documentation seems to be out of date

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -151,46 +162,73 @@ public void handleConnectionChanges(boolean finalPhase) {
...
      * @return A complete expression of this block and all its dependencies.
      */
     public final Expression getFullExpr() {
-        LetExpression fullExpr = new LetExpression(this.getLocalExpr());
-        this.extendExprGraph(fullExpr);
+        Pair<Expression, Set<Block>> localExpr = getLocalExpr();
+        
+        LetExpression fullExpr = new LetExpression(localExpr.a, false);
+        Set<Block> surroundingBlocks = localExpr.b;
+        extendExprGraph(fullExpr, null, surroundingBlocks);
```
wandernauta: I dislike passing a hard `null` here - either note (in the Javadoc, for example) that nulls are fine, or use `Optional`.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -151,46 +162,73 @@ public void handleConnectionChanges(boolean finalPhase) {
...
+    /** Scans for and attaches to a new container, if any */
+    public void refreshContainer() {
+        Bounds myBounds = localToScene(getBoundsInLocal());
+        detachFromContainer();
+        
+        container = parentPane.getBlockContainers().
+            filter(container -> container.localToScene(container.getBoundsInLocal()).contains(myBounds)).
+                reduce((a, b) -> !a.localToScene(a.getBoundsInLocal()).contains(b.localToScene(b.getBoundsInLocal())) ? a : b);
+        
+        container.ifPresent(container -> container.attachBlock(this));
```
wandernauta: Minor nitpick: naming the lambda-local variable the same as the field is a bit confusing.

Code/src/main/java/nl/utwente/viskell/ui/components/BlockContainer.java
```diff
@@ -15,4 +17,25 @@
      */
     public void handleConnectionChanges(boolean finalPhase);
 
+    /** Gets the value of the property boundsInLocal. */
+    public Bounds getBoundsInLocal();
+    
+    /** Transforms a point from the local coordinate space of this Node into the coordinate space of its scene. */
+    public Bounds localToScene(Bounds bounds);
```
wandernauta: Aren't `localToScene` and `getBoundsInLocal` JavaFX methods that every Node gets anyway?

Code/src/main/java/nl/utwente/viskell/ui/components/BlockContainer.java
```diff
@@ -15,4 +17,25 @@
...
+    public void attachBlock(Block block);
+    
+    /** Detach a block from this container */
+    public boolean detachBlock(Block block);
+    
+    /** Detach all blocks from this container */
+    void detachAllBlocks();
+    
+    /** Check whether this container contains the specified block */
+    public boolean containsBlock(Block block);
```
wandernauta: Is this 'contains' as in 'visually falls within the bounds' or 'contains' as in 'logically has a parent-child or grandparent-child relationship'?

Code/src/main/java/nl/utwente/viskell/ui/components/ChoiceBlock.java
```diff
@@ -0,0 +1,140 @@
...
+    protected List<Lane> lanes;
+    
+    /** The output anchor of this block */
+    protected OutputAnchor output;
+    
+    /** The container Node for the Lanes */
+    @FXML protected Pane altSpace;
+    
+    /** The container Node for the OutputAnchor */
+    @FXML protected Pane funSpace;
```
wandernauta: Is `funSpace` the best name here?

Code/src/main/java/nl/utwente/viskell/ui/components/ChoiceBlock.java
```diff
@@ -0,0 +1,140 @@
...
+    @Override
+    public List<OutputAnchor> getAllOutputs() {
+        return Collections.singletonList(output);
+    }
+
+    @Override
+    protected void refreshAnchorTypes() {
+        lanes.stream().forEach(lane -> lane.handleConnectionChanges(false));
+        
+        // TODO make sure the last edited lane gets unified last
```
wandernauta: Why is this important?

Code/src/main/java/nl/utwente/viskell/ui/components/ChoiceBlock.java
```diff
@@ -0,0 +1,140 @@
...
+    public List<Lane> getLanes() {
+        return lanes;
+    }
+    
+    @Override
+    public void relocate(double x, double y) {
+        double dx = x-getLayoutX(), dy = y-getLayoutY();
+        super.relocate(x, y);
+        
+        lanes.forEach(lane -> lane.moveNodes(dx, dy));
```
wandernauta: What's the advantage of making this relative? That would cause the nodes to 'desynchronize' when you miss a relocate. (I'm not sure that can happen, but I imagine it could.)

Code/src/main/java/nl/utwente/viskell/ui/components/Connection.java
```diff
@@ -216,5 +219,18 @@ private void updateBezierControlPoints() {
...
     }
+    
+    /**
+     * Extends the expression graph to include all subexpression required
+     * @param exprGraph the let expression representing the current expression graph
+     * @param container the container to which this expression graph is constrained
+     * @param addLater a mutable list of blocks that have to be added by a surrounding container
+     */
+    protected void extendExprGraph(LetExpression exprGraph, BlockContainer container, Set<Block> addLater) {
+        if ((container == null && !getStartAnchor().block.getContainer().isPresent()) || (container != null && container.containsBlock(getStartAnchor().block)))
```
wandernauta: Missing braces; opportunity to extract variables?

Code/src/main/java/nl/utwente/viskell/ui/components/BlockContainer.java
Flumble: Yes, however BlockContainer can't inherit from Node since that would force it to be a class instead of an interface. Luckily it does use the implementation of Node (or any override down to Lane/LambdaContainer) when requested.

Code/src/main/java/nl/utwente/viskell/ui/components/DefinitionBlock.java
```diff
@@ -147,5 +148,16 @@ public void invalidateVisualState() {
...
+        return body;
+    }
 
+    @Override
+    public void relocate(double x, double y) {
+        double dx = x-getLayoutX(), dy = y-getLayoutY();
+        super.relocate(x, y);
+        
+        body.moveNodes(dx, dy);
+    }
```
wandernauta: This seems to duplicate the method from `ChoiceBlock`. Will every node that has children behave this way?

Code/src/main/java/nl/utwente/viskell/ui/components/InputAnchor.java
```diff
@@ -132,25 +134,32 @@ public void setFreshRequiredType(Type type, TypeScope scope) {
...
-        this.extendExprGraph(expr);
-        return expr;
+        Pair<Expression, Set<Block>> pair = this.getLocalExpr();
+        LetExpression fullExpr = new LetExpression(pair.a, false);
+        Set<Block> surroundingBlocks = pair.b;
+        block.container.ifPresent(container -> extendExprGraph(fullExpr, container, surroundingBlocks));
+        extendExprGraph(fullExpr, null, surroundingBlocks);
+        
+        surroundingBlocks.forEach(block -> block.extendExprGraph(fullExpr, null, new HashSet<>()));
+        return fullExpr;
```
wandernauta: This seems to duplicate `getFullExpr` from `Block`.

Code/src/main/java/nl/utwente/viskell/ui/components/LambdaContainer.java
```diff
@@ -142,16 +149,49 @@ public final void handleConnectionChanges(boolean finalPhase) {
...
+        handleConnectionChanges(false);
+        handleConnectionChanges(true);
+    }
+
+    @Override
+    public boolean detachBlock(Block block) {
+        boolean removed = attachedBlocks.remove(block);
+        block.detachFromContainer();
+        handleConnectionChanges(false);
+        handleConnectionChanges(true);
```
wandernauta: I've seen this (first call with false then call with true) a few times - what does it do? What happens if you forget one of the two steps? Can't the method just do both?

Code/src/main/java/nl/utwente/viskell/ui/components/Lane.java
```diff
@@ -0,0 +1,230 @@
...
+        Pair<Expression, Set<Block>> pair = result.getLocalExpr();
+        LetExpression guards = new LetExpression(pair.a, true);
+        Set<Block> surroundingBlocks = pair.b;
+        result.extendExprGraph(guards, this, surroundingBlocks);
+        
+        attachedBlocks.stream().filter(Block::isBottomMost).forEach(block -> {
+            if (block instanceof MatchBlock) {
+                guards.addLetBinding(((MatchBlock)block).getPrimaryBinder(), block.getAllInputs().get(0).getFullExpr());
+            }
+            else {
```
wandernauta: No newline before `else`.

Code/src/main/java/nl/utwente/viskell/ui/components/BlockContainer.java
Flumble: It's both. Definitely the former, strictly not the latter until containers actually contain nodes instead of having nodes hovering above them.

Code/src/main/java/nl/utwente/viskell/ui/components/ChoiceBlock.java
Flumble: it's in accordance with DefinitionBlock :)

Code/src/main/java/nl/utwente/viskell/ui/components/Lane.java
```diff
@@ -0,0 +1,230 @@
...
+                block.getAllOutputs().forEach(anchor -> {
+                    guards.addLetBinding(new ConstructorBinder("True", Collections.EMPTY_LIST), anchor.getVariable());
```
wandernauta: What does this do?

Code/src/main/java/nl/utwente/viskell/ui/components/ChoiceBlock.java
Flumble: I'll add that

Code/src/main/java/nl/utwente/viskell/ui/components/Lane.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /** Returns the result anchor of this Lane */
+    public ResultAnchor getOutput() {
+        return result;
+    }
+
+    @Override
+    public void refreshAnchorTypes() {
+        // refresh anchor types only once at the start
```
wandernauta: At the start of what?

Code/src/main/java/nl/utwente/viskell/ui/components/Lane.java
```diff
@@ -0,0 +1,230 @@
...
+    @Override
+    public final void handleConnectionChanges(boolean finalPhase) {
+        // avoid doing extra work and infinite recursion
+        if ((!finalPhase && !firstPhaseInProgress) || (finalPhase && !finalPhaseInProgress)) {
+            if (!finalPhase) {
+                // in first phase ensure that anchor types are refreshed if propagating from the outside
+                refreshAnchorTypes();
+
+                firstPhaseInProgress = true;
+                finalPhaseInProgress = false;
```
wandernauta: These variables lok the same at first glance.

Code/src/main/java/nl/utwente/viskell/ui/components/Lane.java
```diff
@@ -0,0 +1,230 @@
...
+    
+    /** Add and initializes a resizer element to this block */
+    private void setupResizer() {
+        resizer = new Polygon();
+        resizer.getPoints().addAll(new Double[]{20.0, 20.0, 20.0, 0.0, 0.0, 20.0});
+        resizer.setFill(Color.BLUE);
+
+        resizer.setManaged(false);
+        this.getChildren().add(resizer);
+        resizer.relocate(300-20, 400-20);
```
wandernauta: This is copied from DefinitionBlock - might be worth extracting or moving to a common superclass.

Code/src/main/java/nl/utwente/viskell/ui/components/Lane.java
```diff
@@ -0,0 +1,230 @@
...
+    public boolean detachBlock(Block block) {
+        if (attachedBlocks.remove(block)) {
+            block.detachFromContainer();
+            handleConnectionChanges(false);
+            handleConnectionChanges(true);
+            block.handleConnectionChanges(false);
+            block.handleConnectionChanges(true);
+            return true;
+        }
+        else {
```
wandernauta: No newline before `else`.

Code/src/main/java/nl/utwente/viskell/ui/components/Pair.java
```diff
@@ -0,0 +1,16 @@
+package nl.utwente.viskell.ui.components;
+
+/**
+ * A generic 2-tuple
+ * @param <A> type of the first element, a
+ * @param <B> type of the second element, b
+ */
+public class Pair<A, B> {
```
wandernauta: There's also `javafx.util.Pair` if you're dead certain you don't want a proper class for this.

Code/src/main/java/nl/utwente/viskell/ui/components/DefinitionBlock.java
Flumble: Hopefully no container will behave this way since it's an artifact of the children not being children in the scene graph.
CBA to extract a class for this one method.

Code/src/main/resources/ui/layout.css
```diff
@@ -127,6 +127,26 @@ Connection {
     -fx-background-radius: 0, 0;
 }
 
+<<<<<<< HEAD
```
wandernauta: Merge gone wrong?

Code/src/main/resources/ui/colours.css
```diff
@@ -156,3 +156,11 @@ ImageView.hoverMenu {
...
     -fx-background-color: transparent;
 }
+
+ChoiceBlock * {
+	-fx-background-color: transparent;
+}
+
+ChoiceBlock #guardSpace {
+	-fx-background-color: rgba(127, 127, 255, 0.3);
+}
```
wandernauta: Newline end of file.

Code/src/main/java/nl/utwente/viskell/ui/components/LambdaContainer.java
Flumble: Good question. Let me refer you to @comonoidial for an explanation what the hell handleConnectionChanges does and why it's in two phases.

Code/src/main/java/nl/utwente/viskell/ui/components/BlockContainer.java
```diff
@@ -15,4 +17,25 @@
      */
     public void handleConnectionChanges(boolean finalPhase);
 
+    /** Gets the value of the property boundsInLocal. */
+    public Bounds getBoundsInLocal();
```
wandernauta: `public` keyword is unnecessary in an interface.

Code/src/main/java/nl/utwente/viskell/ui/components/Pair.java
Flumble: That Pair has the same semantics as Entry: a key-value pair. This is not a key-value pair, but a pair of values.
I agree that components isn't the greatest place for it to reside.

Code/src/main/java/nl/utwente/viskell/ui/components/BlockContainer.java
wandernauta: Okay, that makes sense.

Code/src/main/java/nl/utwente/viskell/ui/components/BlockContainer.java
wandernauta: Understood.

Code/src/main/java/nl/utwente/viskell/ui/components/ChoiceBlock.java
wandernauta: Yes, because the result of a DefinitionBlock is always a function ('hence the fun'). The result of a ChoiceBlock could be any value, right?

Code/src/main/resources/ui/layout.css
Flumble: check

Code/src/main/java/nl/utwente/viskell/ui/components/DefinitionBlock.java
wandernauta: Okay.

Code/src/main/java/nl/utwente/viskell/ui/components/ChoiceBlock.java
Flumble: Since the lanes are going nowhere (relative to it's parent) and I'd like to keep the set of attached nodes private to the Lane it's in, there's no benefit calling moveNodes with absolute coordinates (the lanes will have to subtract the previous position of their parent).
Again, this is an artifact of the current "everything is laid out absolute" mess.

Code/src/main/java/nl/utwente/viskell/ui/components/Lane.java
Flumble: The last line specifically? It tries to bind the evaluation of a function to True, like a normal guard with a function would do, e.g.

    case e of
      a | someBoolExpression -> b

...except `LetExpression`s don't allow for a binding without a left-hand side, so it becomes `True <- someBoolExpression`.

Does it need explaining in the code?

comonoidial:
I consider this "Good Enough:tm:"
Lots of improvements are possible, but not in this PR as it has grown too large already.

PR #300, opened 2015-12-08
Added application block for DefinitionBlocks and

Flumble:
modified expression generation to accommodate for them.

See DefinitionFunction, DefinitionBlock and FunctionBlock for the creation of an application.
The rest is a change from passing required Blocks to required OutputAnchors while generating an expression.

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/haskell/env/FunctionInfo.java
+++ b/Code/src/main/java/nl/utwente/viskell/haskell/env/FunctionInfo.java
@@ -1,6 +1,10 @@
 package nl.utwente.viskell.haskell.env;
 
+import java.util.Collection;
+import java.util.Collections;
+
 import nl.utwente.viskell.haskell.type.Type;
+import nl.utwente.viskell.ui.components.Block;
 
 public abstract class FunctionInfo {
 
@@ -9,6 +13,9 @@ public abstract class FunctionInfo {
     
     /** The type signature the corresponding function. */
     private final Type signature;
+    
+    /** A set of required blocks for this function type */
+    private final Collection<Block> requiredBlocks;
 
     /**
      * @param name The function name.
@@ -17,6 +24,18 @@ public abstract class FunctionInfo {
     public FunctionInfo(String name, Type signature) {
         this.name = name;
         this.signature = signature;
+        this.requiredBlocks = Collections.EMPTY_LIST;
+    }
+
+    /**
+     * @param name The function name.
+     * @param signature The type signature the corresponding function.
+     * @param requiredBlocks A set of blocks required for this function type
+     */
+    protected FunctionInfo(String name, Type signature, Collection<Block> requiredBlocks) {
+        this.name = name;
+        this.signature = signature;
+        this.requiredBlocks = requiredBlocks;
     }
 
     /** @return The internal name of this function. */
@@ -32,6 +51,11 @@ public abstract class FunctionInfo {
     /** @return The a fresh copy of type signature of this function. */
     public final Type getFreshSignature() {
         return this.signature.getFresh();
+    }
+    
+    /** @return The required blocks for this function type */
+    public Collection<Block> getRequiredBlocks() {
+        return requiredBlocks;
     }
 
 }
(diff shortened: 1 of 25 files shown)


Code/src/main/java/nl/utwente/viskell/ui/components/FunctionBlock.java
```diff
@@ -125,17 +126,21 @@ public final Integer getKnotIndex() {
...
         for (InputAnchor in : this.getActiveInputs()) {
-            Pair<Expression, Set<Block>> pair = in.getLocalExpr();
+            Pair<Expression, Set<OutputAnchor>> pair = in.getLocalExpr();
             expr = new Apply(expr, pair.a);
-            addLater.addAll(pair.b);
+            outsideAnchors.addAll(pair.b);
         }
         
-        return new Pair<>(expr, addLater);
+        if (funInfo instanceof DefinitionFunction) {
```
comonoidial: Using instanceof here is a bit ugly, what about adding an extra method to FunctionInfo?

comonoidial:
When building my own map function and applying it to an example it does work when I use the definition application.
But when I use the external output anchor of the definition block instead, I get ghci out of scope errors.
An example of the generated code:
```haskell
let {res__c150b8d = (((:) res__5ab3a4d4) res__1d069c48); lam__6f785057 = (\ map'_x0__2a9f64dd map'_x1__4920cb7e -> let {choiceoutput__2ff98b57 = case (()) of {() | (:) res0__e724b60 res1__3105d392 <- map'_x1__4920cb7e, res__1d069c48 <- ((lam__6f785057 map'_x0__2a9f64dd) res1__3105d392), res__5ab3a4d4 <- ((($) map'_x0__2a9f64dd) res0__e724b60), res__c150b8d <- (((:) res__5ab3a4d4) res__1d069c48), True -> res__c150b8d; () | True -> map'_x1__4920cb7e; }; } in choiceoutput__2ff98b57); val__7e8f969c = ([1..5]); res__2f8c09aa = succ; res__75b021c6 = ((($) lam__6f785057) res__2f8c09aa); res__bc64c73 = ((($) res__75b021c6) val__7e8f969c); } in res__bc64c73
```
Here are in the first let binding res__5ab3a4d4 and res__1d069c48 out of scope.

comonoidial:
It looks like this:
mapexample

Code/src/main/java/nl/utwente/viskell/haskell/env/FunctionInfo.java
```diff
@@ -16,20 +24,38 @@
     public FunctionInfo(String name, Type signature) {
         this.name = name;
         this.signature = signature;
+        this.requiredBlocks = Collections.EMPTY_LIST;
```
wandernauta: Constructors can call other constructors in Java, so this should also work:

    this(name, signature, ImmutableList.of())

Code/src/main/java/nl/utwente/viskell/haskell/env/FunctionInfo.java
```diff
@@ -16,20 +24,38 @@
...
     public final String getName() {
         return this.name;
     }
 
-    /**
-     * @return The a fresh copy of type signature of this function.
-     */
+    /** @return The name of this function used for the front-end. */
+    public String getDisplayName() {
+        return this.name;
```
wandernauta: `return getName()` would perhaps be nicer.

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -153,20 +153,20 @@ public void handleConnectionChanges(boolean finalPhase) {
         }
     }
     
-    /** @return A pair containing the expression this block represents and a set of required blocks. */
-    public abstract Pair<Expression,Set<Block>> getLocalExpr();
+    /** @return A pair containing the expression this block represents and a set of out-of-reach OutputAnchors. */
```
wandernauta: Maybe explain in the Javadoc why getLocalExpr needs to know about out-of-reach OutputAnchors.

wandernauta:
I haven't tested it yet, but this looks neat! :+1:

wandernauta:
Seems to work quite well for the most part. Two questions:

- It appears that changes inside the definition do not (correctly) propagate to usages of the definition outside/do not cause the usages to be re-evaluated: 2015-12-09-130433_726x508_scrot

- There might be a problem with zero-argument functions/'constants': 
2015-12-09-130442_698x491_scrot

Flumble:
Only the display refreshing bug (https://github.com/wandernauta/viskell/pull/300#issuecomment-163203805) is left and occurs in multiple forms (both related and unrelated to this PR).

wandernauta:
Merged, thanks a lot!

PR #303, opened 2015-12-09
Fix removal of lambda block when it is internally connected

comonoidial:


wandernauta:
lgtm :+1:

The fact that you need the `new ArrayList<>` shallow copy is a bit unfortunate, but I can't really think of a better solution.

PR #304, opened 2015-12-09
Reduce size of function menu with a two column block button layout

comonoidial:


wandernauta:
I like this idea. 

Some minor tweaks I would suggest:
- The margins around the buttons could perhaps use some fiddling around with to make them look neater
- I get the extra Close button from a balance standpoint, but I don't think it's good UX - if people see a close button at that position, they will get used to it being there. 'Suddenly' hiding it when the user switches backend would then not be ideal. I'd prefer to just leave the space blank (or purple).
- ...aaand when you're playing around with the FunctionMenu css anyway - maybe the Close button could use a splash of color to help distinguish it from the others?

comonoidial:
Those tweaks make sense, though I don't want to get into UI polishing right now, just got annoying by the growing size of the function menu.

PR #305, opened 2015-12-09
Add the concept of rigid typevars for enforcing polymorpic signatures

### LINKED ISSUE #301: Invalidate type restriction in definition blocks w.r.t their signature
Currently when defining a definition block with a type variable, the result type can be restricted to a concrete type (or have additional type classes).
However, a result type should be *at least as generic* as the definition's signature.

A possible fix would be to check if the type feeding into the result anchor does not add restrictions to the signature's result type when unified.

Example of an offending type restriction:
shot_1226

comonoidial:
This resolves #301

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/haskell/type/TypeVar.java
+++ b/Code/src/main/java/nl/utwente/viskell/haskell/type/TypeVar.java
@@ -15,7 +15,7 @@ public class TypeVar extends Type {
     /**
      * An optional mutable reference to a concrete type.
      */
-    public final static class TypeInstance {
+    protected final static class TypeInstance {
         /**
          * The textual representation of the type variable.
          */
@@ -26,6 +26,11 @@ public class TypeVar extends Type {
          */
         private final boolean internal;
 
+        /**
+         * Whether this type variable is rigid (meaning that it can't be unified with a more specific or concrete type)
+         */
+        private boolean isRigid;
+        
         /**
          * A concrete instantiated type or null.
          */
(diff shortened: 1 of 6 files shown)


Code/src/main/java/nl/utwente/viskell/haskell/type/TypeVar.java
```diff
@@ -79,23 +98,42 @@ private ConcreteType get() {
...
 
         /**
          * Deeply unifying all aspects of type instances
          * @param the other type instance.
-         * @throws HaskellTypeError if the combined constraint set of the typeva instances is not satisfiable.
+         * @throws HaskellTypeError if the combined constraint set of the typevar instances is not satisfiable.
+         * @throws HaskellTypeError if a rigid type variable gets unified with more constraints or another rigid type variable
          */
         private void unifyWith(TypeInstance other) throws HaskellTypeError {
+            if (this == other) {
```
Flumble: Does it have to be the same reference or an isomorphic type?

comonoidial: Same reference as that is the exception to the condition 4 lines further.

Flumble: If both are rigid and isomorphic, they shouldn't produce an exception, right? A rigid ::a and ::b can live happily together.

comonoidial: I don't think so, the type a -> a is not the same as the type a -> b

PR #307, opened 2015-12-14
Improved vertical currying and added binary operator blocks

comonoidial:


Flumble:
Looks great, except for (partially) curried binary functions where the text goes through the result space.
shot_1228

Also, wouldn't it be nicer if all infix functions were infixed instead of only binary ones? (I'm looking at you, `.`!)

PR #310, opened 2015-12-16
A container for every block and related refactorings

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/components/LambdaContainer.java
```diff
@@ -171,43 +174,32 @@ public void invalidateVisualState() {
...
-        attachedBlocks.forEach(node -> node.relocate(node.getLayoutX()+dx, node.getLayoutY()+dy));
-    }
-
-    /** Remove all associations of this container with others in preparation of removal, including all connections */
-    public void removeAllLinks() {
+    public void deleteAllLinks() {
        this.args.forEach(OutputAnchor::removeConnections);
        this.res.removeConnections();
-       this.detachAllBlocks();
+       this.attachedBlocks.forEach(block -> block.moveIntoContainer(this.wrapper.getPane().getTopLevel()));
```
Flumble: Shouldn't the blocks move into the parent container (or the wrapper's parent)?

PR #314, opened 2015-12-16
Limit the size of value and display blocks

comonoidial:


wandernauta:
Looks good to me, thanks!

PR #316, opened 2015-12-17
Reorganized catalog, balancing catagories and split off floating point

comonoidial:


wandernauta:
This makes a lot of sense, thanks!

PR #318, opened 2015-12-18
Give an indication of type in widths of wires

comonoidial:

PR #319, opened 2015-12-19
Show scoping errors on connections

comonoidial:
Depends on #318


wandernauta:
This seems to work well, and I like the idea a lot.

I don't know if it's related, but in some (apparently random) cases moving a block inside a lambda makes its connections show scope errors even when they shouldn't. Moving the block again fixes them.

2015-12-21-113255_446x286_scrot

comonoidial:
I think this is because sometime a block gets related to the wrong container. Now it is obvious when it happens we should fix those bugs.

PR #322, opened 2015-12-19
Fix two small bugs

comonoidial:
One is avoiding a concurrent modifications exceptions on deleting a lambda or case.
Second is stopping a warning from breaking evaluator value parsing.


wandernauta:
Would it make sense to disable all warnings (`-w`) until we have a way of properly dealing with them?

comonoidial:
Maybe, but I think this is the only one showing up feeding expressions to ghci.

wandernauta:
Merged, thanks!

PR #323, opened 2015-12-20
Support copying of simple blocks from the context menu

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/components/BinOpApplyBlock.java
```diff
@@ -215,7 +216,16 @@ private void dragShiftOuput(double y) {
 		return ImmutableList.of(this.output);
 	}
 
-	@Override
+    @Override
+    public Optional<Block> getNewCopy() {
+        if (this.leftInput.curried || this.rightInput.curried) {
+            return Optional.empty();
```
wandernauta: Why can't we make a copy in this case?

wandernauta:
It's not ideal that copy doesn't work on more complex cases, but this is a good start and it seems to work well.

:+1:

PR #330, opened 2015-12-21
Size and positioning tweaks of function menu

comonoidial:


Code/src/main/resources/catalog/haskell.xml
```diff
@@ -192,7 +192,7 @@
 
 
   <functions>
-    <category name="Comparison and logic">
+    <category name="Conditionals">
```
Flumble: This catagory contains exactly comparison operators and boolean logic. Conditionals seems unnecessarily vague unless you put guards or other things in there.

PR #336, opened 2015-12-22
Make sure that a container block is not hidden behind its parent

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/CustomUIPane.java
```diff
@@ -377,14 +377,30 @@ public void clearChildren() {
     }
 
     public Stream<BlockContainer> getBlockContainers() {
-        return bottomLayer.getChildrenUnmodifiable().stream().flatMap(node -> {
-            if (node instanceof ChoiceBlock) {
-                return ((ChoiceBlock)node).getLanes().stream();
-            }
-            if (node instanceof DefinitionBlock) {
-                return Stream.of(((DefinitionBlock)node).getBody());
+        return bottomLayer.getChildrenUnmodifiable().stream().flatMap(node ->
```
Flumble: nitpick: a filter in between would be nicer here

Code/src/main/java/nl/utwente/viskell/ui/components/Block.java
```diff
@@ -227,6 +227,10 @@ public void moveIntoContainer(BlockContainer target) {
             this.container = target;
             target.attachBlock(this);
             
+            if (this.getInternalContainers().size() > 0) {
```
Flumble: nitpick: testing on empty is nicer here

PR #338, opened 2015-12-23
More reliable function menu spawning with touch

comonoidial:
Tested on surface.


wandernauta:
Does this need to be tested on the big screen as well? It has a different number of dots per inch, so I expect [this](https://github.com/wandernauta/viskell/pull/338/files#diff-43386581ab8d05bb47dea7bb51e21e20R238) value might have to be tweaked.

wandernauta:
As it turns out, it doesn't need to be. Will merge, thanks!

PR #339, opened 2015-12-23
Add opening and closing animation to the function menu

comonoidial:
Also prevent accidental use during that.


wandernauta:
Hmm. I'm not sure I see the advantage here. The animation doesn't really serve a purpose; it doesn't make it any clearer what is going on.

A hypothetical professional Viskell programmer would open the function menu dozens of times during a nine-to-five working day. I think any animation would get old after seeing it so many times.

comonoidial:
It's purpose is to make the UI feel responsive, while adding delays to prevent accidental touch actions.
And for now I care mainly about the upcoming demo, and find speculative arguments about the long term not convincing.

comonoidial:
Now the animations are skipped on mouse actions

wandernauta:
Thanks!

PR #340, opened 2015-12-23
Make slider blocks more usefull and add an integer variant

comonoidial:
Also the new sliders allow negative numbers, and the scaling is relative and exponential.


wandernauta:
Yep, this seems to work quite well. Thanks! :+1:

PR #342, opened 2015-12-24
Moved things around so that the UIPane doesn't have to be global anymore

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/MainOverlay.java
```diff
@@ -111,4 +126,38 @@ public void setTouchVisible(boolean value) {
...
+        double scale = this.mainPane.getScaleX();
+
+        /* Limit zoom to reasonable range. */
+        if (scale <= 0.2 && ratio < 1) return;
+        if (scale >= 3 && ratio > 1) return;
+
+        this.mainPane.setScaleX(scale * ratio);
+        this.mainPane.setScaleY(scale * ratio);
+        this.mainPane.setTranslateX(this.mainPane.getTranslateX() * ratio);
+        this.mainPane.setTranslateY(this.mainPane.getTranslateY() * ratio);
```
wandernauta: This seems a little feature envy-ish (only ever accesses and changes fields on `ToplevelPane`, doesn't even touch `MainOverlay`)

wandernauta:
lgtm :+1:

PR #346, opened 2015-12-25
Use the actual area of blocks for the clickable region

comonoidial:
By default panes use their outside bounds to determine whether they are the target of a click, but for blocks with all these anchors sticking out those bounds are too large.


wandernauta:
Works great! Thanks.

PR #349, opened 2015-12-27
Add block for arbitrary values based on output type using QuickCheck

comonoidial:


Code/src/main/java/nl/utwente/viskell/ghcj/GhciEvaluator.java
```diff
@@ -20,7 +20,7 @@ public GhciEvaluator() throws HaskellException {
 
     @Override
     protected List<String> getModules() {
-        return ImmutableList.of("Data.List", "Data.Maybe", "Data.Either", "Data.Monoid", "Control.Applicative");
+        return ImmutableList.of("Data.List", "Data.Maybe", "Data.Either", "Data.Monoid", "Control.Applicative", "Test.QuickCheck");
```
wandernauta: It seems that, while `Test.QuickCheck` is a part of Haskell Platform, it's not distributed as part of the GHC you get from the Ubuntu repositories (the other ones are).

I don't really want to make this a preference - you wouldn't be able to open Viskell programs made in an environment where QC is available in one where it isn't.

Can we bundle QuickCheck somehow?

Code/src/main/java/nl/utwente/viskell/ui/components/ArbitraryBlock.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private void getNextValue(int seed) {
+        this.setValue("?????");
+        
+        if (! this.output.hasConnection()) {
+            return;
+        }
+
+        // we cannot generate values for complicated types
```
wandernauta: Can't QuickCheck make values for every type that's a member of `Arbitrary`?

Code/src/main/java/nl/utwente/viskell/ui/components/ArbitraryBlock.java
```diff
@@ -0,0 +1,87 @@
...
+
+            public void onFailure(Throwable throwable) {
+                Platform.runLater(() -> ArbitraryBlock.this.setValue("....."));
+            }
+        });
+    }
+    
+    @Override
+    public void invalidateVisualState() {
+        if (this.value.getText().startsWith("?")) {
```
wandernauta: I don't really like this check - can we use a boolean or something?

Code/src/main/java/nl/utwente/viskell/ui/components/ArbitraryBlock.java
comonoidial: If we have a polymorph type like 'Arbitary a => [a]', what values could be generated for that? You don't know enough of 'a' yet.

Code/src/main/java/nl/utwente/viskell/ghcj/GhciEvaluator.java
comonoidial: QuickCheck is very common Haskell library, and if someone doesn't have it they can just install it using cabal or stack.

wandernauta: We'd probably want to warn the user, though, instead of failing silently, right?

Code/src/main/java/nl/utwente/viskell/ui/components/ArbitraryBlock.java
wandernauta: Well, you could generate `[]` I guess... but I can see why it wouldn't.

Code/src/main/java/nl/utwente/viskell/ghcj/GhciEvaluator.java
comonoidial: Would it be doable to catch the ghci error messages on startup and show that to the user?

wandernauta: Yep, that sounds like an easy fix - if ghci fails to start, tell the user to install ghci and quickcheck.

PR #350, opened 2015-12-27
Avoid animation delay on menus activated by mouse

comonoidial:


comonoidial:
Merging as it is a fairly trivial convenience change.

PR #351, opened 2015-12-27
Make drawing wire easier and some added visual polish

comonoidial:


comonoidial:
Just tested it on surface, drawing connections with touch seems a lot easier now.

PR #352, opened 2015-12-27
Resize containers to fit blocks when they are dropped into them

comonoidial:


comonoidial:
This is the first step in automatic resizing behaviour, a lot more could be done, but it would be nice to get some experience with this first.

PR #370, opened 2015-12-31
Inititate only one ghci session on startup and test QuickCheck evaluation

comonoidial:


Code/src/main/java/nl/utwente/viskell/ghcj/GhciSession.java
```diff
@@ -60,6 +60,14 @@ public GhciSession() {
...
         queue = new ArrayBlockingQueue<>(1024);
         errors = EvictingQueue.create(LOG_SIZE);
+        switch (pickBackend()) {
+            case Clash: 
+                this.catalog = new HaskellCatalog("/catalog/clash.xml");
+                break;
+            default:
+                this.catalog = new HaskellCatalog("/catalog/haskell.xml");
+                break;
+        }
```
wandernauta: I liked the old setup better, of course - I'd like GhciSession to have as little knowledge of its backends as possible. Maybe make `getCatalogPath` static?

Code/src/main/java/nl/utwente/viskell/ui/Main.java
```diff
@@ -61,6 +54,35 @@ public void start(Stage stage) throws Exception {
...
+                        Platform.runLater(() -> new Alert(Alert.AlertType.WARNING, msg).showAndWait());
+                    }
+                }
+            });
+        } catch (RuntimeException e) {
+            String msg = "It seems the Glasgow Haskell Compiler, GHC, is not " +
+                    "available. Executing programs will not be enabled. We " +
+                    "strongly recommend you install GHC, for example by " +
+                    "installing the Haskell Platform (haskell.org/platform)." +
+                    "Or it might be that the QuickCheck is not installed.";
```
wandernauta: Could use some copyediting.

wandernauta:
Other than the minor points above, this looks good. I like pulling the construction of the catalog out of the asynchronous startup - lots of bits and pieces need the catalog in some way or other.

PR #376, opened 2016-01-01
Experiment with giving wires subtle color variations

comonoidial:


wandernauta:
Can you add a screenshot of what this is supposed to look like? It might just be because it's, as you say, subtle, but I'm only seeing black wires.

comonoidial:
I now think we should try other things first for making complex programs more readable, making things more colorful is just some workaround with minor effects

PR #381, opened 2016-01-03
Rewrite of lambdas and local definitions, with various bug fixed added in.

comonoidial:


comonoidial:
Because of the large conflict potential and important bug fixes, this is going in early.

PR #391, opened 2016-01-04
Stabilize function block knot by not centering the block

### LINKED ISSUE #390: Flickering when currying horizontally-curried applications
When currying functions with the 'Vertical curry' preference turned off, the width of the FunctionBlock varies ('flickers') whenever the mouse is moved.

This seems to be introduced by 98a30f80565fa15486511e69a01392728fb35ebf.

comonoidial:
Fixes #390.


wandernauta:
Seems to work! :+1:

PR #392, opened 2016-01-04
Fix touch on categories, and some button cosmetics

comonoidial:


comonoidial:
Tested on the big screen.

PR #394, opened 2016-01-04
Make sliders less sensitive and added reset

comonoidial:


Code/src/main/java/nl/utwente/viskell/ui/CircleMenu.java
```diff
@@ -79,7 +79,13 @@ private CircleMenu(Block block, boolean byMouse) {
             this.add(copy);
         }
 
-        if (block instanceof LambdaBlock) {
+        if (block instanceof SliderBlock) {
```
wandernauta: Looking at CircleMenu in its current state it seems like it could be much improved by having blocks know which buttons their menu should have (i.e. a `getContextButtons` on Block). That would remove the need for all the instanceof checks and casts.

PR #397, opened 2016-01-04
Tweak the visuals of the lambda block, more lambda is better

comonoidial:


wandernauta:
λooks good to me. :+1:

PR #410, opened 2016-01-06
Polish the visuals of various details

comonoidial:


Flumble:
What's the rationale for rounding all function blocks and rounding the top side of match blocks?

PR #422, opened 2016-01-13
Gesture for quickly cutting a connection wire

comonoidial:
A connection can be cut by drawing a line across it and then tracing that line back to its starting point.
This can be done either with as a single finger drag or dragging with the primary mouse button.
Also the panning by mouse has been moved to a secondary button.


wandernauta:
lgtm :+1:

PR #426, opened 2016-03-25
Use a TopMenu on Mac OS X instead of onscreen context menu for touch devices.

andrewdavidmackenzie:
Added MacTopMenu to create a Top application menu for Mac.
GlobalMenu refactored into GlobalContextMenu that is used for on-screen context menu, and MenuActions which are shared between all menus.
MainOverlay now wraps overlay in a BorderPane that accepts a TopMenu, and when detected we are running on a Mac it replaces the on-screen
context menu with a TopMenu that does the same actions (except full-screen which is a window control in Mac OS X).

Avoid referencing overlay via Main.java public overflow, use the parameter passed in to the constructor and pass down to refreshTheme()
MainOverlay doesn't need to be public in Main.java any more.

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/GlobalContextMenu.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/GlobalContextMenu.java
@@ -1,0 +1,28 @@
+package nl.utwente.viskell.ui;
+
+import javafx.scene.control.ContextMenu;
+import javafx.scene.control.MenuItem;
+
+/**
+ * A context menu with global actions (i.e. quit).
+ */
+public class GlobalContextMenu extends ContextMenu {
+    public GlobalContextMenu(MenuActions menuActions) {
+        super();
+
+        MenuItem menuPreferences = new MenuItem("Preferences...");
+        menuPreferences.setOnAction(menuActions::showPreferences);
+
+        MenuItem menuInspector = new MenuItem("Inspector");
+        menuInspector.setOnAction(menuActions::showInspector);
+
+        MenuItem menuFullScreen = new MenuItem("Toggle full screen");
+        menuFullScreen.setOnAction(menuActions::toggleFullScreen);
+        
+        MenuItem menuQuit = new MenuItem("Quit");
+        menuQuit.setOnAction(menuActions::onQuit);
+
+        this.getItems().addAll(menuActions.fileMenuItems());
+        this.getItems().addAll(menuInspector, menuPreferences, menuFullScreen, menuQuit);
+    }
+}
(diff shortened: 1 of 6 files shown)


andrewdavidmackenzie:
This is a small visual change as a way of making a first contact with the codebase.

As I develop and use on Mac OS X I thought I'd make this small change to make it look and act more like a standard app on a desktop OS.

I wanted to use the standard "App" menu instead of having the "java" menu, and put preferences and quit there as normal Mac OS X apps, but that requires some additional libraries (JavaFX doesn't allow it out-of-the-box) and so I stopped.

BTW: I'm looking at doing importing of blocks from saved JSON (for importing a saved file initially, maybe other uses later). Currently battling gson on that (as the export seems a bit not standard) and will let you know.

wandernauta:
This looks great code-wise! :+1: Thanks a lot.

I don't have a Mac handy to see how it looks, though - could you paste a screenshot?

wandernauta:
As the person... ahem... responsible for the exporter, I can confirm that it doesn't really follow any standard, and it's probably incomplete besides. I think it really depends on whether we want to stick to the 'every element can be serialized to and from a map of strings' assumption.

If we do, any file format that stores these maps will do, and we're not really tied to gson or json; one could just replace `Exporter.export` wholesale. 

If we don't want to keep that assumption, we're not really tied to gson either, and it could be a good time to think about a better file format in general. I haven't really looked into how other visual languages handle this. (Maybe we could write annotated SVG files or something equally strange?)

andrewdavidmackenzie:
screen shot 2016-03-26 at 13 16 31

wandernauta:
Fantastic! Again, thank you very much.

andrewdavidmackenzie:
On the export/import, I have something starting to work (problems with Connections/Wires between loaded Blocks).... - maybe best if I open an issue (maybe a WIP PR later) to discuss separately. Although here is an advance.

You could use GSON totally, but that leads to full classnames and "automatic" reconstruction that I don't think would work in this case (as well as needing default constructors for all objects).

I've gone the route of adding more data to the export, splitting it into two collections of two different types (Blocks and Connections) and then using GSON to parse that into two Maps.

I have added a static fromBundleFragment() method to all objects subject to deserialization - that take care of the per-class specifics (in the "properties" submap) and instantiating an object of the type needed.

The Block superclass has a fromBundle() method that finds the class based on the short classname ("kind") in the JSON, finding the fromBundleFragment() method for that class and invoking it....hence getting an instantiated subclass of Block object created....then I add it to the TopLevelPanel (preserving x and y)...

I keep track of loaded blocks and their IDs (from the export, not the current ID) in a temporary map.

Then, I parse the connections, finding the Blocks and Anchors involved from the temporary map, and wiring things up (trying at the moment).

wandernauta:
That sounds entirely reasonable. I'm looking forward to the PR!

PR #428, opened 2016-03-26
Attempt at implementing write (serialization) and read (deserialization) of content

andrewdavidmackenzie:
NOTE: Type (signature) serialization/deserialization has not been implemented yet.

See https://github.com/viskell/viskell/issues/155


Code/src/main/java/nl/utwente/viskell/haskell/env/CatalogFunction.java
```diff
@@ -33,6 +37,31 @@
...
+    @Override
+    public Map<String, Object> toBundleFragment() {
+        Map<String, Object> bundleFragment = new HashMap<>();
+        bundleFragment.put(Bundleable.KIND, this.getClass().getSimpleName());
+        bundleFragment.put("name", name);
+        bundleFragment.put("signature", signature.toString());
+        bundleFragment.put("category", category);
+        bundleFragment.put("documentation", documentation);
+        bundleFragment.put("isConstructor", isConstructor);
+        bundleFragment.put("isCommon", isCommon);
```
comonoidial: Why not save the name only and then lookup the corresponding function info from the catalog?

comonoidial:
Looks good overall :+1: 
As for the serialization of types I would just use strings for now.

I wonder what a good way of testing this functionality is.
Having some json files (of non trivial examples) in the testsuite which are loaded and saved back again would work. But having to manually fixup the json files after every significant changes is annoying.

By the way I see a lot of unrelated code style improvements in this PR. These are fine, but to make the diffs easier to read I would prefer them in a seperate PR in the future.

Code/src/main/java/nl/utwente/viskell/haskell/env/CatalogFunction.java
andrewdavidmackenzie: I did't know enough about how all this works to for that idea to even occur to me.
If it can be done, then great and would simplify this serialization a lot.

Any example code to lookup a CatalogFunction from the name and instantiate it, would be appreciated.

andrewdavidmackenzie:
Thanks!

# Types
There are a number of types of "Types", and I'm not sure how to construct instances of them on deserialization. Suggestions would be great and I can add to the PR.

# Testing
Yes, testing this type of thing can be tricky.

At the unit test level, I would add unit tests for all classes supporting Bundleable (mocking collaborators) where the test creates an instance and tests it can serialise to bundle (toBundleFragment etc).... then test a new instance can be created from such a bundle....and *maybe* check equivalence.

Checking equivalence might mean adding "equals" code to many classes - just for testing....so not sure the best....or just have the test assert equals on a number of fields it knows about.

Some of the code structure is not so conducive to unit testing. e.g. Just creating a Connection adds it to the ToplevelPane in the constructor. It would be better to reduce that coupling and do less in constructors, being able to create one (for testing) independantly - and then the app would add it to the ToplevelPane separately. You can see how ToplevelPane needs to be passed to all the deserialization methods.....

I can take a shot at some Unit Tests in the PR, or in a later PR if you want....Let me know.

Integration testing is tougher - maybe programatically create a ToplevelPane and add blocks and connections to it (making sure the serialized content keeps track of changes to the classes involved), and then serializing and checking all is there OK.....and/or deserializing and checking all is the same.

Having tests based on saved files is a good way to check you're not making any changes that break backwards compatibility....when breaking changes are "accepted", then you need to update the saved files and document the change. Maintaining backwards compatibility for serialized objects is a pain always....Maybe wait for Viskell v1.0.0 and then add a version number to the file, and then *try* to keep compatibility for all v1.*.* versions.

# Style changes
Agreed. Sorry about that. They crept in as I was trying to keep files I modified free of warnings in IDE....will keep separate in the future.

wandernauta:
As for backwards compatibility, we have the rare advantage that there are literally zero Viskell files floating around right now - putting a version (or perhaps a file format revision?) number should be enough, I think.

Code/src/main/java/nl/utwente/viskell/haskell/env/CatalogFunction.java
comonoidial: Seems like the HaskellCatalog is lacking a method for that, so the ugly way to do this now is:
` toplevelPane.getGhciSession().getCatalog().getByPredicate(fn -> fn.getName().equals(name)).iterator().next();`

wandernauta: Wouldn't `toplevelPane.getEnvInstance().lookupFun(name)` also work?

comonoidial: Right, that is much better.

comonoidial:
I think all components (such as ConstantBlock) that have types that need serialization are from user input, thus saving them as plain strings would be good enough for now.

Unit testing this codebase is not really practical, as the graphical elements and programming language bits are very much tied together.

As everything is still in an experimental stage, just take the approach you think that would take least effort now and to keep up in the near future.
It's fine for the testing to come in a later PR and it doesn't have to be complete.

andrewdavidmackenzie:
I took a look at doing some unit tests, see this file as an example that tests ToplevelPane serialization/deserialization as an example:

https://github.com/andrewdavidmackenzie/viskell/blob/master/Code/src/test/java/nl/utwente/viskell/ui/ToplevelPaneTest.java

Using powermock, you should be able to get around a lot of the problems - but it's hard work!

To try get my whenNew() test working with powermock I spent a day in dependency hell, then gave up :-(

To make testing much easier the recommended way would be to use dependency injection (e.g. Dagger or similar) to avoid all those "new MyClass()" object instantiations in the code and have them injected when running, and mocks injected when testing.

# Next Steps?
What to do now?
Add to the PR:
- Add suggested code for adding a function from the catalog?
- Need to decide how to serialize/deserialize types -> help welcome
- A few working tests from my example above (more as an example than extensive test coverage)?

andrewdavidmackenzie:
Are you guys Ok with the proposed next steps?

Add to the PR:
- Add suggested code for adding a function from the catalog using it's name
- Need to decide how to serialize/deserialize types -> *help needed here*
- A few working tests using mocking to illustrate how to do some, and issues with others

If so, I will work on later this week/weekend....

andrewdavidmackenzie:
BTW: I have no idea what the "QuantumRenderer" test failure is about, but other travis failures related to PowerMock I will have to look at.

wandernauta:
I don't know enough about PowerMock's specific brand of magic to say anything reasonable about the test failures, but the proposed steps are fine by me and the code looks great :+1: 

I believe Quantum is JavaFX's rendering layer, and the problem in this case might be that the fake screen we hook up to Travis (`xvfb`) isn't good enough:

    Prism-ES2 Error : GL_VERSION (major.minor) = 1.4

OpenGL 1.4 is positively ancient.

    Xlib:  extension "RANDR" missing on display ":99.0".

Xrandr is multi-screen support, I can imagine JavaFX wanting that and xvfb not supplying it.

    Graphics Device initialization failed for :  es2, sw

You'd expect 'sw' (which I think might stand for 'software') to work, but...

    Error initializing QuantumRenderer: no suitable pipeline found

...apparently it doesn't.

Finally, it could be that rendering to the virtual framebuffer was always broken and we just never noticed.

andrewdavidmackenzie:
Do any of the tests actually do that!?

It does switch workspaces (as if the UI was starting) when I execute tests
I think.  Trying to start the UI during unit tests should be
avoided/avoidable.

wandernauta:
Well, you can't test a component without constructing it. At the moment you can't construct a component if you don't have a ToplevelPane. You can't make a ToplevelPane if you don't have JavaFX ready, and you can't have JavaFX ready without doing the initialization dance. The JavaFX initialization dance needs something to draw to, so that's why we have to start UI.

The tests responsible (`test/java/nl/utwente/viskell/ui/components/*.java`) don't seem to test anything useful, though, so if the easiest way forward is to rip them out, I'm all for it.

andrewdavidmackenzie:
I guess it depends on if you want to unit test an object, or integration
test it together with JavaFX.

For a real unit test we'd mock all collaborating objects to avoid starting
JavaFX

Looking at those tests I think you could just mock(ToplevelPane.class) like
I do with Pane.class and other FX objects here:

https://github.com/andrewdavidmackenzie/viskell/blob/master/Code/src/test/java/nl/utwente/viskell/ui/ToplevelPaneTest.java

That test instantiates a ToplevelPane (FX) object and mocks collaborating
ones (such as Pane) - but doesn't inherit from ComponentTest and doesn't
start FX.

Andrew

Code/src/main/java/nl/utwente/viskell/haskell/env/CatalogFunction.java
andrewdavidmackenzie: lookupFun(9 returns FunctionInfo not a CatalogFunction, so I guess I'll have to make a new constructor for CatalogFunction() that accepts a FunctionInfo - I need to read the code more...

wandernauta: Hmm. It could be that some components reference CatalogFunctions where a FunctionInfo would also suffice, which would make this a bit easier. I don't have the tools right now to check that, though.

comonoidial: AFAIK all FuntionInfos in Environment are CatalogFunctions now, thus the types in Environment could be restricted for now. How to deal with a library of user defined functions is a problem for later.

comonoidial:
If Travis doesn't support testing things involving UI code, I would be fine with having a seperate build target for such tests that can only be run manually.
You have spent a lot time on testing issues already, and there are a lot more interesting things to built and experiment with.

andrewdavidmackenzie:
Hi Guys, OK, taken a second pass at this....

* I have code (locally) that removes the test dependancy on starting FX that I mentioned, but I ran into the same problem with powermock and junit. Raised the issue with them, and they have an issue in github to fix it I'm tracking. When they fix that we can then take a look at testing again, without changing all the code structure.

* It now serialized with a version number and checks it on opening a file.
   (no nice error dialogs yet....)

* It shows the saved filename as the window title.

* added a samples directory with trivial samples of primitives for testing, plus some simple samples. Only the observe_a_number.viskell sample works due to the problem mentioned below....

* I am stuck on deserializing the Types. See last commit and FIXME comments in ConstantBlock.java.
 Easy to reproduce opening the sample/hello_world.viskell file.
I think the best would be to merge this as is (it does more than before) and for us to look at it together. WDYT?

If we can merge this, then I will rebase and merge some menu improvements (shortcuts) and propose to you some areas I'm interested in exploring... Let me know.

andrewdavidmackenzie:
I don't have access to a Linux machine to debug those Travis failures :-( I'm not sure if they are reproducible locally on a linux box, but I assume so.

There are solutions described here http://stackoverflow.com/questions/21185156/javafx-on-linux-is-showing-a-graphics-device-initialization-failed-for-es2-s

Code/src/main/java/nl/utwente/viskell/ghcj/GhciSession.java
```diff
@@ -37,7 +37,7 @@
     private Evaluator ghci;
 
     /** Gets filled with a HaskellCatalog instance when ghci is ready. */
-    private HaskellCatalog catalog;
+    private static HaskellCatalog catalog;
```
wandernauta: Are you sure this is right? I think that there could potentially be multiple GhciSessions with different catalogs.

comonoidial:
I have added a workaround for deserializing ConstantBlocks in: https://github.com/viskell/viskell/tree/pr/428
It seems like github does not support pushing directly to a PR thus it became a new branch instead.

Code/src/main/java/nl/utwente/viskell/haskell/env/FunctionInfo.java
```diff
@@ -1,14 +1,28 @@
...
+import java.util.HashMap;
+import java.util.Map;
 
 public abstract class FunctionInfo {
 
+    private static final Map<String, String> subClassMap;
+    static {
+        Map<String, String> aMap = new HashMap<>();
+        aMap.put(CatalogFunction.class.getSimpleName(), CatalogFunction.class.getName());
+        subClassMap = Collections.unmodifiableMap(aMap);
```
wandernauta: I haven't tested it, but this could perhaps be an `ImmutableMap.of` which would be slightly shorter.

Code/src/main/java/nl/utwente/viskell/haskell/env/FunctionInfo.java
```diff
@@ -19,6 +33,19 @@ protected FunctionInfo(String name, Type signature) {
...
+    public static FunctionInfo fromBundleFragment(Map<String,Object> bundleFragment) throws NoSuchMethodException, InvocationTargetException, IllegalAccessException, ClassNotFoundException {
+        String kind = (String)bundleFragment.get(Bundleable.KIND);
+        String className = subClassMap.get(kind);
+        Class<?> clazz = Class.forName(className);
+
+        // Find the static "fromBundleFragment" method for the named type and call it
+        Method fromBundleMethod = clazz.getDeclaredMethod("fromBundleFragment", Map.class);
+        return (FunctionInfo)fromBundleMethod.invoke(null, bundleFragment);
+    }
+
```
wandernauta: Hmm.

I believe (*believe*) the only FunctionInfo's that aren't from the catalog come from your own definitions, and that's just because those don't really have a category. CatalogFunction is the only subclass, and really the only special thing about catalog functions is that they have some extra strings (description, category). If we need to do this kind of magic to keep that distinction alive I think we might be better off collapsing the hierarchy, hoisting the extra strings up into FunctionInfo and making them optional (or even `Optional`).

Code/src/main/java/nl/utwente/viskell/ui/components/ApplyAnchor.java
```diff
@@ -35,7 +39,19 @@ public ApplyAnchor(int baseArity) {
...
-        
+
+    public Map<String, Object> toBundleFragment() {
+        return ImmutableMap.of(
+                Bundleable.KIND, this.getClass().getSimpleName(),
+                "arity", this.baseArity
+        );
+    }
+
+    public static ApplyAnchor fromBundleFragment(Map<String, Object> bundleFragment) throws ClassNotFoundException, NoSuchMethodException, IllegalAccessException, InvocationTargetException {
```
wandernauta: In what scenario would this throw these exceptions?

Code/src/main/java/nl/utwente/viskell/ui/serialize/ViskellFormat.java
```diff
@@ -0,0 +1,42 @@
+package nl.utwente.viskell.ui.serialize;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Created by andrew on 03/04/16.
```
wandernauta: We don't do these kind of file headers elsewhere - just as easy to pull this information from `git blame`.

I'll make a PR to add your name to LICENSE, though.

Code/src/test/java/nl/utwente/viskell/ui/ToplevelPaneTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
+import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.whenNew;
+
+/**
+ * Unit tests for Serialization/Deserialization from ToplevelPane
+ * Created by andrew on 27/03/16.
```
wandernauta: See above - might be a IDE template thingy, I think.

Code/src/test/java/nl/utwente/viskell/ui/ToplevelPaneTest.java
```diff
@@ -0,0 +1,202 @@
...
+        @SuppressWarnings("MismatchedQueryAndUpdateOfCollection")
+        List<Map<String, Object>> connections = new ArrayList<>();
+        connectionsBundle.put(ToplevelPane.CONNECTIONS_SERIALIZED_NAME, connections);
+
+        // UUT
+        ToplevelPane toplevelPane = new ToplevelPane(mockGhci);
+
+        // Test
+        toplevelPane.fromBundle(connectionsBundle); // no crash
+    }
```
wandernauta: This is great :+1:

Code/src/test/java/nl/utwente/viskell/ui/ToplevelPaneTest.java
```diff
@@ -0,0 +1,202 @@
...
+         Reason: javassist.bytecode"
+
+
+    Spent about a day in junit/powermock/javassist dependency hell trying many different versions
+    of junit and powermock that produce different failures I couldn't solve either....such as
+            "java.lang.VerifyError: Inconsistent stackmap frames at branch target 1135"
+            (that one seems to be Java 7 related and can be avoided passing "-noverify" to JVM when running tests,
+            but that seems a hack)
+
+    ....then I gave up...
```
wandernauta: I get:

```
org.mockito.exceptions.misusing.NotAMockException: 
Argument passed to verify() is of type ToplevelPane and is not a mock!
Make sure you place the parenthesis correctly!
See the examples of correct verifications:
    verify(mock).someMethod();
    verify(mock, times(10)).someMethod();
    verify(mock, atLeastOnce()).someMethod();

at nl.utwente.viskell.ui.ToplevelPaneTest.addsBlockFromBundleFromBundleTest(ToplevelPaneTest.java:199)
```

wandernauta:
I do have a Linux box, and everything seems to work just fine - loading/saving works, and all tests pass (except for the ignored one). The code looks great. I'd be fine with merging as is, unless @comonoidial has any objections or you want to tackle some of my comments beforehand.

comonoidial:
I would like to see the Travis failures resolved in some way before merging this.
The code is ok for merging, though improving on the easily addressable comments would be nice.

Code/src/main/java/nl/utwente/viskell/ghcj/GhciSession.java
andrewdavidmackenzie: If that's the case, then we will need bigger changes.... as we need to get "a"/"the" catalog from the static fromBundleFragment() method way down the call stack....so the deserialization (starting at toplevel pane or menu actions) will need to know *which* session to use to deserialize with, and pass it down all the fromBundle() & fromBundleFragment() calls to where it is needed.

Code/src/main/java/nl/utwente/viskell/ui/components/ApplyAnchor.java
andrewdavidmackenzie: You're right. Looks like a copy and paste error.

Code/src/main/java/nl/utwente/viskell/ui/serialize/ViskellFormat.java
andrewdavidmackenzie: Comes from IDE template. I'll remove. Not worried about attribution.

Code/src/test/java/nl/utwente/viskell/ui/ToplevelPaneTest.java
andrewdavidmackenzie: Yep. removing.

Code/src/test/java/nl/utwente/viskell/ui/ToplevelPaneTest.java
andrewdavidmackenzie: I think that's because the mocking transformation of the file (byte code manipulation) fails...

andrewdavidmackenzie:
Not sure why it works for you and not travis then.... it must be some dependency that is not captured and installed when the travis tests execute?

Code/src/main/java/nl/utwente/viskell/ghcj/GhciSession.java
comonoidial: I think it save to assume for now that only one active ghci session exist as you have only one active toplevel working area. Leave those bigger changes for when they are really needed.

andrewdavidmackenzie:
@wandernauta if you could try and reproduce the Travis failure locally we could try and fix it...I will take a look also in case not Linux specific....

wandernauta:
@comonoidial experimented with getting Travis to work with JavaFX
yesterday, but I don't believe he got it running. I'm more and more
thinking that doing UI integration tests on a headless environment like
Travis is not ideal.

I have sinced managed to reproduce the problem with the JavaFX backend not
starting - both the hardware-accelerated and software renderers seem to
work as expected when run on a real screen, a real graphics card and a real
keyboard, but fail in half a dozen different ways when run under Xvfb.
Meanwhile, real headless testing for JavaFX seems to have been delayed
until Java 9 (http://stackoverflow.com/a/12230114).

Maven has a distinction between tests (test) and integration tests
(verify). Maybe we could tell Travis to run the first, and run the second
ourselves? It's not ideal, but it beats compiling our own Java to include
the experimental headless rendering backend.

comonoidial:
I agree that splitting off the UI tests in Maven for manual execution is the best workaround for now.

andrewdavidmackenzie:
http://antoniogoncalves.org/2012/12/13/lets-turn-integration-tests-with-maven-to-a-first-class-citizen/

wandernauta:
That looks like a reasonable way to do it. Changes required would be:

- add IT suffix to tests that hit UI in some way
- add failsafe to POM
- change .travis.yml to run test instead of verify

I don't think we need the additional magic to optionally skip either UT or IT.

andrewdavidmackenzie:
Do you prefer to do that, or want me to modify the PR with those changes - until Travis passes?

wandernauta:
I'd prefer the latter, if that's okay with you - I think it's cleanest if we merge the PR once it's green.

andrewdavidmackenzie:
http://i.ebayimg.com/images/a/(KGrHqF,!mEE+7-iS4k3BQPGI+fp9g~~/s-l300.jpg

wandernauta:
Fantastic! I'll take a look at merging this soon.

andrewdavidmackenzie:
Then we can create some issues for remaining problems and unimplemented things in the import.

wandernauta:
Merged! Thanks a lot.

PR #429, opened 2016-03-26
New resets current file

andrewdavidmackenzie:
Before this change if you saved a file, then cleared everything with “New” then saved again - it would overwrite the first file.

Now the current file is reset on “New” and the user has to choose where to save.


wandernauta:
Good catch! Cherry-picked into master. Thanks!

PR #440, opened 2016-04-10
Add handling of scroll events to the MainOverlay - this makes right-click scrolling (two finger drag) work on Mac OS X.

andrewdavidmackenzie:
Can someone not on a Mac make sure this doesn’t cause double scrolling for other OS - as it seems it already worked on other OS?


comonoidial:
It seems to work both with mouse and touchpad under windows, but for a touchscreen it needs an extra side condition: `&& !event.isDirect()` to avoid mixing up with other events.
For panning the event handling should not be done on the MainOverlay but on the ToplevelPane which uses the TouchContext helper class for event handling.

comonoidial:
Replaced by #441.

Issue #439, opened 2016-04-10
Zoom bar has disapeared


andrewdavidmackenzie:
Seems like some changes I did to get menus have removed the zoom bar 

Fix it, and get it back.

comonoidial:
Resolved by #442.

PR #441, opened 2016-04-11
Support panning using a touchpad or scrollwheel

comonoidial:
This could use some testing on non windows devices.
Replaces PR #440.

Unfortunately I could not get zooming to work on touchpad too, and I don't know whether to blame the OS or JavaFX for that.

### DIFF
--- a/Code/src/main/java/nl/utwente/viskell/ui/FunctionMenu.java
+++ b/Code/src/main/java/nl/utwente/viskell/ui/FunctionMenu.java
@@ -69,11 +69,6 @@ public class FunctionMenu extends StackPane implements ComponentLoader {
         this.parent = pane;
         this.loadFXML("FunctionMenu");
         this.dragContext = new DragContext(this);
-        /*
-         * Consume scroll events to prevent mixing of zooming and list
-         * scrolling.
-         */
-        this.addEventHandler(ScrollEvent.SCROLL, Event::consume);
 
         /* Create content for searchSpace. */
         // TODO create search tools and add them to searchSpace.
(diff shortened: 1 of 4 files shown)


wandernauta:
Seems to work on Linux - I can pan in both directions with two-finger touchpad scroll.

Would it make sense to add a 'reverse scroll direction' preference option? It feels a bit more natural to me reversed, but I think that would very much be a personal preference thing.

TheMSB:
> wandernauta

I think adding a preference option for reverse scrolling is a good idea that would require much extra work to implement. Some people just really prefer scrolling one way or the other so it could help with the usability of the program.

comonoidial:
I will add a preference. What is the commonly agreed default on this?

wandernauta:
I think the direction you're using in this PR (positive scroll) is a reasonable default.

andrewdavidmackenzie:
At least on Mac, the reverse scroll thing is an OS setting, and doesn't
need handing by the app

andrewdavidmackenzie:
I have tested on Mac.
- works fine, direction is the direction set by the OS
- I checked that if you switch the OS scroll behaviour it switches also, so no further changes needed there

I notice that if you have the cursor on top of a popped-up toolbar, then the scrolling doesn't work....I assume because it's on top of the layer that where you are processing the scroll events....the toolbar layer must consume them.
To fix I guess you'd need to handle scroll events on the top (touch?) layer, OR, make sure the toolbar doesn't consume scroll events and they fall through to the lower layers.

In my branch for scrolling, I did the initial displacement of the background (as you suggested) so it is harder to get to see the white border. Maybe could do too?
A better fix, would be if we had scrolled to the edge of the canvas and not allow scrolling beyond it...so you can never see the white.

Same thing for zoom out, I thought if we zoom out beyond the size of the canvas, then trancate position to extent of canvas and avoid seeing the white around it.

comonoidial:
The toolbar had a overreaching eventhandler consuming scroll event, which is fixed in last commit.

I didn't change displacement now, because whether that is needed depends on the resolution of #439.

Limiting the scrolling to the edge of the canvas is a separate issue, which has multiple possible design alternatives.

andrewdavidmackenzie:
+1 Fine to merge for me then.

Issue #444, opened 2017-04-18
Don't fork processes in the constructor of Evaluator


travisfw:
A constructor should be as free of side effects as possible. If I/O has to be done before other operations, I would make this initialization explicit with, for example, a ```void init()``` method. I would make the constructor protected and provide a static factory method that also does the initialization and I/O, throwing an IOException or HaskellIOException. A HaskellIOException which subclasses IOException instead of just [wrapping the IOException in a HaskellException](https://github.com/viskell/viskell/blob/8af529873e6aaf0dfa6ff477611fc71d3b01456f/Code/src/main/java/nl/utwente/viskell/ghcj/Evaluator.java#L60 ) would make methods that perform I/O more obvious, and I/O exceptions could be handled in the outside context intelligently, whether HaskellIOException or other.

Forking processes is an even heavier operation than I/O, and it should always be transparent to the calling context that it is happening. You can accidentally create zombies and other hazards on the host machine when you fork new system processes. Definitely not something to do in a constructor.

I just cloned viskell and first did ```mvn package``` and the tests requiring haskell executables [failed at the constructor to Evaluator](https://github.com/viskell/viskell/blob/8af529873e6aaf0dfa6ff477611fc71d3b01456f/Code/src/test/java/nl/utwente/viskell/ghcj/GhciEvaluatorTest.java#L16 ).

```stacktrace
Test set: nl.utwente.viskell.ghcj.GhciEvaluatorTest
-------------------------------------------------------------------------------
Tests run: 3, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 0.016 sec <<< FAILURE!
typeErrorTest(nl.utwente.viskell.ghcj.GhciEvaluatorTest)  Time elapsed: 0.01 sec  <<< ERROR!
nl.utwente.viskell.ghcj.HaskellException: java.io.IOException: Cannot run program "stack": error=2, No such file or directory
        at nl.utwente.viskell.ghcj.Evaluator.<init>(Evaluator.java:60)
        at nl.utwente.viskell.ghcj.GhciEvaluator.<init>(GhciEvaluator.java:13)
        at nl.utwente.viskell.ghcj.GhciEvaluatorTest.startGhci(GhciEvaluatorTest.java:16)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #445, opened 2017-04-18
Skip tests that depend on Haskell binaries on the host machine.

travisfw:
I limited my edits to just the test classes, but also filed the bug #444 earlier.

Ideally, the test for whether Haskell is installed would NOT involve an attempt to fork a processes, which it currently does. But changing that would require deeper refactoring. I'm happy to do that, but I'd like your feedback first. And I haven't even run Viskell yet; I'm just getting past the tests.


comonoidial:
We never considered the case of running the tests without having haskell installed.
Thanks for fixing this oversight.

